PR #41, opened 2015-11-19
Make the deb_repo parameter to both GCE and AWS packer templates opti…

duftler:
…onal.

Make install_packages.sh script tolerant of missing deb_repo.
Comment-out debianRepository attribute from checked-in rosco.yml, but leave it with some comments as an example.

Tested:
  Amazon bake with package from specified deb_repo.
  Amazon bake with redis-server and no deb_repo specified.
  Google bake with package from specified deb_repo.
  Google bake with redis-server and no deb_repo specified.


duftler:
@skim1420 please review.

skim1420:
LGTM

PR #42, opened 2015-11-19
removed aws os: ubuntu definitions from rosco.yml. name "ubuntu" is a…

skorten:
…mbiguous and amis were for 12.04

This is a fix for https://github.com/spinnaker/spinnaker/issues/597

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -45,29 +45,6 @@ aws:
     # TODO(duftler): Support additional store types beyond ebs.
     # TODO(duftler): Support additional operating systems.
     # TODO(duftler): Feels like just about everything here could be (optionally) specified per bake.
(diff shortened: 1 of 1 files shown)


duftler:
The root of the issue is that the Bake stage configuration ui has a hardcoded set of choices. If we remove the config in this pr, the user can still select 'ubuntu' in the ui and then the bake would fail. I think the complete fix is to have the ui query the bakery for the list of supported images and then dynamically render that list of choices.

ttomsu:
> the complete fix is to have the ui query the bakery for the list of supported images and then dynamically render that list of choices.

This is what I'm working on for https://github.com/spinnaker/spinnaker/issues/597

skorten:
Ah, that makes sense and I agree is a better solution.  I will close this PR.

PR #43, opened 2015-11-20
Set initial interval of rush health ping to 5 minutes.

duftler:
Will help avoid the initial stack-trace when rush and rosco are starting up at the same time.
@spinnaker/google-reviewers please review.


duftler:
Or @spinnaker/netflix-reviewers .

PR #45, opened 2015-11-23
Add support for AWS associate_public_ip_address property (default to …

duftler:
…true).

Add support for GCE use_internal_ip property (default to false).
Rename AWSBakerDefaults fields subnetId and vpcId to awsSubnetId and awsSubnetId.
Closes https://github.com/spinnaker/spinnaker/issues/576.
@dpeterka please review.


dpeterka:
This looks good to me.   Pondering if we shouldn't make this more configurable over time.

duftler:
"More configurable" in what way? Can you provide more details please?

Adding each option one-at-at-time as folks request them doesn't seem like a great plan.

Btw, somewhat related, @ttomsu is working on making the bake stage ui more dynamic based on the configuration of the bakery itself. That will make it much easier to control what base images are available. Maybe following that work we can figure out how to surface additional template-specific options as well?

PR #46, opened 2015-11-23
Use CANCELLED everywhere instead of CANCELED.

duftler:
This is necessary since orca/MonitorBakeTask expects CANCELLED.
@spinnaker/reviewers please review.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeStatus.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeStatus.groovy
@@ -46,7 +46,7 @@ class BakeStatus implements Serializable {
   String resource_id
 
   static enum State {
-    PENDING, RUNNING, COMPLETED, SUSPENDED, CANCELED
+    PENDING, RUNNING, COMPLETED, SUSPENDED, CANCELLED
   }
 
   static enum Result {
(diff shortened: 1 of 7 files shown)


ajordens:
Generally speaking we've opt'd to go with the American spelling throughout Spinnaker (_says the guy who tried to introduce Judgement_).

IMO, I would suggest normalizing `CANCELLED` to `CANCELED` in `MonitorBakeTask`. It doesn't look like `BakeStatus` is used outside of that class, and nothing is specifically referencing `CANCELLED`.

duftler:
That sounds good to me.

The main reason I elected to go with what `MonitorBakeTask` expected was because I assumed that was a consequence of what your internal bakery returns here: https://github.com/spinnaker/orca/blob/master/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/BakeryService.groovy#L35 (and on line 32 just above).

duftler:
@ajordens Did you get a chance to eyeball `MonitorBakeTask` to see if this change makes sense?

ajordens:
I still think we should support both CANCELLED and CANCELED in the MonitorBakeTask.

It's true that our internal bakery speaks CANCELLED but I think it's worth being consistent (as far as American spellings go) on the components and code we're responsible for (as much as that pains me to say as a non-US speller).

duftler:
Closing in favor of https://github.com/spinnaker/orca/pull/660.

PR #53, opened 2015-11-25
Treat packer config files as dpkg configuration files.

duftler:
This will result in them surviving package uninstallation/upgrade.
User will be prompted by installed if there is a conflict and the old files will be renamed to install_packages.sh.dpkg-old.
@dpeterka please review.
@layered-esell fyi


duftler:
Thanks for showing me how to get this to work @dpeterka .

dpeterka:
Looks good did we want to consider adding perhaps everything under /opt/spinnaker/conf/* ?

duftler:
Closes https://github.com/spinnaker/spinnaker/issues/625.

duftler:
Good idea. I can do that in a separate PR under `spinnaker/spinnaker`.

duftler:
Wonder if `configurationFile()` supports a wildcard?

dpeterka:
documentation does not mention if it does or does not I would assume it's just going to append the string to a text file so I don't see why not. Issue then would be if .deb supports it.

duftler:
Cool, will try that next. Going to merge this and kick off a release so I can try the full upgrade path via apt-get.

PR #56, opened 2015-11-28
First step in exposing customizable Bake options: exposing them in Rosco

duftler:
Cherry-pick @ttomsu's commit from #48.
Needed rebasing but I don't have write permissions to the forked repo.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeOptions.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeOptions.groovy
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2015 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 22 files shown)


duftler:
Closes #48.

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -50,7 +53,27 @@ class BakeryController {
...
-  @RequestMapping(value = '/{region}/bake', method = RequestMethod.POST)
+  @RequestMapping(value = '/bakeOptions', method = RequestMethod.GET)
+  List<BakeOptions> bakeOptions() {
+    cloudProviderBakeHandlerRegistry.list().collect { it.getBakeOptions() }
+  }
+
+  @RequestMapping(value = '/bakeOptions/{cloudProvider}', method = RequestMethod.GET)
+  BakeOptions bakeOptionsByCloudProvider(@PathVariable("cloudProvider") BakeRequest.CloudProviderType cloudProvider) {
+    def bakeHandler = cloudProviderBakeHandlerRegistry.lookup(cloudProvider)
+    if (!bakeHandler) {
```
duftler: @ttomsu The param is an enum, so the only way this can be true is if the enum is updated to support a new cloud provider, but the cloud provider does not register itself.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeOptions.groovy
```diff
@@ -0,0 +1,39 @@
...
+  List<BaseImage> baseImages
+
+  static class BaseImage {
+    String id
+    String shortDescription
+    String detailedDescription
+    BakeRequest.PackageType packageType
+  }
+
+  static class Selected {
```
duftler: Is this necessary? We can't just hold a reference to `BakeOptions.BaseImage` directly in places where `BakeOptions.Selected` is now used?

ttomsu:
Issue tag: https://github.com/spinnaker/spinnaker/issues/597

PR #59, opened 2015-12-14
Scope incomplete bakes to specific Rosco instance.

duftler:
With this change, each Rosco instance will only query for status updates for its own incomplete bakes.


duftler:
@ttomsu please review.

PR #60, opened 2015-12-15
Adds detailed view of base Images to be used in Orca

ttomsu:
Orca needs to be able to dynamically figure out the packageType for the specified base OS. This PR exposes the baseImage object endpoint.

@duftler @lwander PTAL


ttomsu:
Tag: https://github.com/spinnaker/spinnaker/issues/597

duftler:
LGTM.

PR #63, opened 2016-01-21
Add support to both AWS and GCE for overriding base image via base_am…

duftler:
…i bake request attribute.
Add support to AWS for enhanced_networking bake request attribute.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
@@ -54,7 +54,13 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
 
     // TODO(duftler): Work through definition of uniqueness.
     bakeRequest.with {
-      return "bake:$cloud_provider_type:$region:$vm_type:$base_os:$package_name"
+      def enhancedNetworkingSegment = bakeRequest.enhanced_networking ? 'enhancedNWEnabled' : 'enhancedNWDisabled'
+
+      if (bakeRequest.base_ami) {
+        return "bake:$cloud_provider_type:$region:$vm_type:$base_os:$bakeRequest.base_ami:$enhancedNetworkingSegment:$package_name"
+      } else {
+        return "bake:$cloud_provider_type:$region:$vm_type:$base_os:$enhancedNetworkingSegment:$package_name"
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


duftler:
@spinnaker/reviewers please review.
This relates to https://github.com/spinnaker/spinnaker/issues/683, but does not close it out completely (the naming attributes are not support yet).

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -53,8 +53,13 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
     }
 
     // TODO(duftler): Work through definition of uniqueness.
+    // TODO(duftler): Include aws_associate_public_ip_address and enhanced_networking in key?
```
cfieber: aws_associate_public_ip_address is a packer setting on how it connects to the instance in order to do snapshotting when producing the AMI, it has not effect on the resulting image

enhanced_networking should be in the key because that is an attribute of the resulting image

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/config/RoscoAWSConfiguration.groovy
```diff
@@ -63,6 +65,7 @@ class RoscoAWSConfiguration {
     List<AWSVirtualizationSettings> virtualizationSettings = []
   }
 
+  @AutoClone(style = AutoCloneStyle.SIMPLE)
```
cfieber: an alternative to this is `@Immutable(copyWith=true)` and you could do
`virtualizationSettings.copyWith(sourceImage: bakeRequest.base_ami)`

cfieber:
LGTM

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
duftler: Done.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/config/RoscoAWSConfiguration.groovy
duftler: Tried (and failed) to use `@Immutable`. Wasn't able to figure out how to convince spring to initialize the beans from the properties without the fields of `GCEVirtualizationSettings` and `AWSVirtualizationSettings` ending up `null`. Guessing it relates to the way I'm initializing `GCEBakeryDefaults` and `AWSBakeryDefaults`.

PR #67, opened 2016-02-18
Fix for issue #739, tag packer build instances

peterpd:


duftler:
Thanks for the contribution.
Would using appversion be more useful as the run-tag? It has the package(s) encoded, as well as the job and commit details.
Here's an example of an appversion tag: appversion: hello-karyon-rxnetty-1-h9.011821b624b4dfad4c4a985ba100a15948cf4d44/TestJob2/9

It is exploited in the ui to populate some of the "Package" details: 
image

peterpd:
Hi Matt,
Do you like to use appversion, instead of packages?

Or is it OK to add both appversion and packages?

Thanks,

Peter

peterpd:
Hi Matt,

Since appversion has a lot more info, let me use appversion instead.

Thanks,

Peter

peterpd:
Hi Matt,

Tried with appversion, but the value is not populated.

Added this line to aws-ebs.json
"run_tags": {"appversion": "{{user `appversion`}}"}

Anything else needed to have appversion populated?

Thanks,
Peter

duftler:
Ah, sorry, realized that this value only gets populated if the pipeline was
triggered from Jenkins and the Jenkins job details are in the context. Ok,
packages it is then.

Hi Matt,

Tried with appversion, but the value is not populated.

Added this line to aws-ebs.json
"run_tags": {"appversion": "{{user appversion}}"}

Anything else needed to have appversion populated?

Thanks,
Peter

—
Reply to this email directly or view it on GitHub
<https://github.com/spinnaker/rosco/pull/67#issuecomment-185926293>.

peterpd:
Thanks Matt. I'll leave the change as is then.

duftler:
Lgtm.

PR #68, opened 2016-02-19
upgrade support

gardleopard:
Run upgrades on the ami if you send upgrade with value true to api.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
@@ -55,6 +55,7 @@ class BakeRequest {
   Boolean enhanced_networking
   String ami_name
   String ami_suffix
+  Boolean upgrade
 
   @ApiModelProperty("The explicit packer template to use, instead of resolving one from rosco's configuration")
   String template_file_name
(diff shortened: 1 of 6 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
```diff
@@ -55,6 +55,7 @@ class BakeRequest {
   Boolean enhanced_networking
   String ami_name
   String ami_suffix
+  Boolean upgrade=false
```
duftler: Can drop the `=false` here and guard the parameterMap instead. This will be more consistent with the way other parameters are handled.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -120,6 +120,8 @@ abstract class CloudProviderBakeHandler {
 
     parameterMap.configDir = configDir
 
+    parameterMap.put("upgrade", bakeRequest.upgrade)
```
duftler: Please handle the setting of this new parameter the same way as `bakeRequest.build_host` just above.

rosco-web/config/packer/install_packages.sh
```diff
@@ -16,4 +17,7 @@ if [[ "$deb_repo" != "" ]]; then
 fi
 
 sudo apt-get update
+if [[ "$upgrade" == "true" ]]; then
```
duftler: Please surround this conditional block with blank lines to keep the formatting consistent.

rosco-web/config/packer/aws-ebs.json
```diff
@@ -15,6 +15,7 @@
     "build_host": "",
     "deb_repo": "",
     "packages": "",
+    "upgrade": "",
```
duftler: Please make the same changes to `gce.json` since it also employs `install_packages.sh`.
Trying to keep the behavior of the two templates consistent where reasonable to do so.

duftler:
LGTM.

PR #72, opened 2016-02-24
Added ami_name and ami_suffix to the bakeKey

kalymero:


duftler:
Reading through the PR, I see some opportunity to factor out a bit of the common (cross-provider) code into the base/abstract class. I'll put up a PR for you to have a look at later today that does this, and I think it'll make your changes here even easier to apply (and alleviate some of the chore of making the same changes in multiple bake handlers).
Then, if we agree, we can merge the base PR and then rebase this one on top of it, with just the remaining changes.
Sound good?

kalymero:
Hi Matt
sounds very good.  I'll wait for your PR then.  Thanks

duftler:
Here is the PR I mentioned: https://github.com/spinnaker/rosco/pull/73
Should be straightforward to rebase your changes in this PR on that one (I will have to merge that one first).
If you make your changes to add `ami_name` and `ami_suffix` in just `CloudProviderBakeHandler.produceBakeKey()` after this PR is merged, it will apply to all providers.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -52,16 +57,35 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
...
       } else {
-        return "bake:$cloud_provider_type:$region:$vm_type:$base_os:$enhancedNetworkingSegment:$package_name"
+        name = package_name
       }
+
+      def suffix
+      if (bakeRequest.ami_suffix) {
+        suffix = bakeRequest.ami_suffix
+      } else {
+        suffix = clock.millis()
```
duftler: I know you'll be reworking this PR anyway, but wanted to mention that including the timestamp here will cause the bake key to not work as intended. If the bake key is used to determine if a bake being requested has already been requested, it has to be possible to generate the same bake key on subsequent calls.

duftler:
LGTM.

PR #76, opened 2016-03-03
Fail the build if there is an error installing packages

gardleopard:
We have seen failing package installs with amazon linux reported as successful bakes.

This makes the install packages script fail on errors.

### DIFF
--- a/rosco-web/config/packer/install_packages.sh
+++ b/rosco-web/config/packer/install_packages.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Make rush fail the build on errors.
+set -e
+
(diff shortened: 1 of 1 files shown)


rosco-web/config/packer/install_packages.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e # Make rush fail the build
```
duftler: Please move the comment to the preceding line and terminate it with a period (trying to match the other comments in this script).

gardleopard:
@duftler I fixed it and squashed the commits.

duftler:
Thanks for the fix.
LGTM.

PR #78, opened 2016-03-10
Dont start services when installing packages on debian based machines

gardleopard:
When services are started on install, sometimes the bake fails because its unable to umount the disk:
` Stderr: umount: /mnt/packer-amazon-chroot-volumes/xvdg/dev: device is busy.`

https://www.packer.io/docs/builders/amazon-chroot.html look at gotchas at the end.

The reason this is done in install_packages.sh and not in the template is because of support for non debian based distros.

### DIFF
--- a/rosco-web/config/packer/install_packages.sh
+++ b/rosco-web/config/packer/install_packages.sh
@@ -18,6 +18,14 @@ packages=`echo $packages | sed 's/^"\(.*\)"$/\1/' | sed 's/,/ /g'`
 
 
 function provision_deb() {
+  # https://www.packer.io/docs/builders/amazon-chroot.html look at gotchas at the end.
+  if [[ "$disable_services" == "true" ]]; then
+    echo "creating /usr/sbin/policy-rc.d to prevent services from being started"
+    echo '#!/bin/sh' > /usr/sbin/policy-rc.d
+    echo 'exit 101' >> /usr/sbin/policy-rc.d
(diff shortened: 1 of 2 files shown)


rosco-web/config/packer/install_packages.sh
```diff
@@ -18,6 +18,12 @@ packages=`echo $packages | sed 's/^"\(.*\)"$/\1/' | sed 's/,/ /g'`
 
 
 function provision_deb() {
+  # https://www.packer.io/docs/builders/amazon-chroot.html look at gotchas at the end
```
duftler: Please use terminating period for comment.

rosco-web/config/packer/install_packages.sh
```diff
@@ -29,6 +35,10 @@ function provision_deb() {
 
   # Enforce the package installation order.
   for package in $packages; do sudo apt-get install --force-yes -y $package; done
+
+  # https://www.packer.io/docs/builders/amazon-chroot.html look at gotchas at the end
```
duftler: Please use terminating period for comment.

duftler:
Since this is specifically a workaround for the aws-chroot builder, we should probably only enable this behavior if a flag is passed from the template. Maybe a new (optional) `disable_services` flag?

gardleopard:
@duftler I have added it as a flag in the chroot template now. Also I have added period after the comments.

duftler:
LGTM.
Thanks!

PR #79, opened 2016-03-19
Use sudo when running aws-chroot builder

gardleopard:
This makes packer run as sudo if it is a aws-chroot build. There is a corresponding pull request to the spinnaker repo to add the spinnaker user to sudoers https://github.com/spinnaker/spinnaker/pull/816.

mock script to test with:
```
#!/bin/bash
echo "Running mock"
env
echo "sleep"
sleep 5
echo "done sleeping"
echo "i am"
whoami
echo "sleep"
sleep 30
```

output of mock script:
```
Running mock
SHELL=/bin/sh
TERM=linux
USER=root
SUDO_USER=spinnaker
SUDO_UID=1001
USERNAME=root
MAIL=/var/mail/root
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
SHLVL=1
SUDO_COMMAND=/usr/local/bin/packer build -color=false -var aws_region=us-west-2 -var aws_ssh_username=ubuntu -var aws_instance_type=t2.micro -var aws_source_ami=ami-3d50120d -var aws_target_ami=trusty-vim-gard-2-all-1458387928895-trusty -var aws_associate_public_ip_address=true -var package_type=deb -var packages=vim -var configDir=/opt/rosco/config/packer /opt/rosco/config/packer/aws-chroot.json
HOME=/home/spinnaker
LOGNAME=root
SUDO_GID=1001
_=/usr/bin/env
sleep
done sleeping
i am
root
sleep
```

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
@@ -42,6 +42,9 @@ abstract class CloudProviderBakeHandler {
   @Value('${yumRepository:}')
   String yumRepository
 
+  @Value('${templatesNeedingRoot:}')
+  List<String> templatesNeedingRoot
+
   /**
    * @return A cloud provider-specific set of defaults.
    */
@@ -120,7 +123,11 @@ abstract class CloudProviderBakeHandler {
    * Returns the command that should be prepended to the shell command passed to the container.
    */
   String getBaseCommand() {
-    return ""
+    if (templatesNeedingRoot) {
+      return templatesNeedingRoot.contains(getBakeryDefaults().templateFile) ? "sudo" : ""
+    } else {
+      return ""
+    }
   }
 
   /**
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
@@ -20,8 +20,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
 
   @Override
   List<String> buildPackerCommand(String baseCommand, Map<String, String> parameterMap, String absoluteTemplateFilePath) {
-    def packerCommand = ["packer", "build", "-color=false"]
-
+    def packerCommand = [baseCommand, "packer", "build", "-color=false"]
     parameterMap.each { key, value ->
       if (key && value) {
         def keyValuePair = value instanceof String && value.contains(" ") ? "$key=\"$value\"" : "$key=$value"
@@ -32,6 +31,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
     }
 
     packerCommand << absoluteTemplateFilePath
+    packerCommand.removeAll([null, ""])
 
     packerCommand
   }
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.rosco.providers.aws
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.rosco.api.Bake
 import com.netflix.spinnaker.rosco.api.BakeRequest
+import com.netflix.spinnaker.rosco.config.RoscoConfiguration
 import com.netflix.spinnaker.rosco.providers.aws.config.RoscoAWSConfiguration
 import com.netflix.spinnaker.rosco.providers.util.ImageNameFactory
 import com.netflix.spinnaker.rosco.providers.util.PackerCommandFactory
@@ -43,10 +44,11 @@ class AWSBakeHandlerSpec extends Specification {
   @Shared
   RoscoAWSConfiguration.AWSBakeryDefaults awsBakeryDefaults
 
+  @Shared
+  RoscoConfiguration roscoConfiguration
+
   void setupSpec() {
     def awsBakeryDefaultsJson = [
-      awsAccessKey: "FOO",
-      awsSecretKey: "BAR",
       templateFile: "aws_template.json",
       defaultVirtualizationType: "hvm",
       baseImages: [
@@ -287,8 +289,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "t2.micro",
@@ -326,8 +326,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-amzn"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ec2-user",
         aws_instance_type: "t2.micro",
@@ -354,6 +352,67 @@ class AWSBakeHandlerSpec extends Specification {
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, "$configDir/$awsBakeryDefaults.templateFile")
   }
 
+  void 'produces packer command with all required parameters for amzn, with sudo'() {
+    setup:
+      def awsBakeryDefaultsJson = [
+        templateFile: "aws-chroot.json",
+        defaultVirtualizationType: "hvm",
+        baseImages: [
+          [
+            baseImage: [
+              id: "amzn",
+              packageType: "RPM",
+            ],
+            virtualizationSettings: [
+              [
+                region: REGION,
+                virtualizationType: "hvm",
+                instanceType: "t2.micro",
+                sourceAmi: SOURCE_AMZN_HVM_IMAGE_NAME,
+                sshUserName: "ec2-user"
+              ]
+            ]
+          ]
+        ]
+      ]
+      RoscoAWSConfiguration.AWSBakeryDefaults localAwsBakeryDefaults = new ObjectMapper().convertValue(awsBakeryDefaultsJson, RoscoAWSConfiguration.AWSBakeryDefaults)
+
+      def imageNameFactoryMock = Mock(ImageNameFactory)
+      def packerCommandFactoryMock = Mock(PackerCommandFactory)
+      def bakeRequest = new BakeRequest(user: "someuser@gmail.com",
+        package_name: PACKAGE_NAME,
+        base_os: "amzn",
+        vm_type: BakeRequest.VmType.hvm,
+        cloud_provider_type: BakeRequest.CloudProviderType.aws)
+      def targetImageName = "kato-x8664-timestamp-amzn"
+      def parameterMap = [
+        aws_region: REGION,
+        aws_ssh_username: "ec2-user",
+        aws_instance_type: "t2.micro",
+        aws_source_ami: SOURCE_AMZN_HVM_IMAGE_NAME,
+        aws_target_ami: targetImageName,
+        repository: YUM_REPOSITORY,
+        package_type: BakeRequest.PackageType.RPM.packageType,
+        packages: PACKAGE_NAME,
+        configDir: configDir
+      ]
+
+      @Subject
+      AWSBakeHandler awsBakeHandler = new AWSBakeHandler(configDir: configDir,
+        awsBakeryDefaults: localAwsBakeryDefaults,
+        imageNameFactory: imageNameFactoryMock,
+        packerCommandFactory: packerCommandFactoryMock,
+        templatesNeedingRoot: [ "aws-chroot.json" ],
+        yumRepository: YUM_REPOSITORY)
+
+    when:
+      awsBakeHandler.producePackerCommand(REGION, bakeRequest)
+
+    then:
+      1 * imageNameFactoryMock.deriveImageNameAndAppVersion(bakeRequest, _) >> [targetImageName, null, PACKAGE_NAME]
+      1 * packerCommandFactoryMock.buildPackerCommand("sudo", parameterMap, "$configDir/aws-chroot.json")
+  }
+
   void 'produces packer command with all required parameters for ubuntu, using default vm type, and overriding base ami'() {
     setup:
       def imageNameFactoryMock = Mock(ImageNameFactory)
@@ -366,8 +425,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "t2.micro",
@@ -405,8 +462,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "m3.medium",
@@ -445,8 +500,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         template_file_name: "somePackerTemplate.json")
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "m3.medium",
@@ -485,8 +538,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         extended_attributes: [someAttr1: "someValue1", someAttr2: "someValue2"])
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "m3.medium",
@@ -526,8 +577,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-trusty"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "t2.micro",
@@ -569,8 +618,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         cloud_provider_type: BakeRequest.CloudProviderType.aws)
       def targetImageName = "kato-x8664-timestamp-trusty"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "t2.micro",
@@ -612,8 +659,6 @@ class AWSBakeHandlerSpec extends Specification {
                                         upgrade: true)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
       def parameterMap = [
-        aws_access_key: awsBakeryDefaults.awsAccessKey,
-        aws_secret_key: awsBakeryDefaults.awsSecretKey,
         aws_region: REGION,
         aws_ssh_username: "ubuntu",
         aws_instance_type: "t2.micro",
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
@@ -1,0 +1,29 @@
+package com.netflix.spinnaker.rosco.providers.util
+
+import spock.lang.Shared
+import spock.lang.Specification
+
+class LocalJobFriendlyPackerCommandFactorySpec extends Specification {
+
+  @Shared
+  LocalJobFriendlyPackerCommandFactory packerCommandFactory = new LocalJobFriendlyPackerCommandFactory()
+
+  void "packerCommand handles baseCommand as null, empty string and real string"() {
+    setup:
+    def parameterMap = [
+      something:   something
+    ]
+
+    when:
+    def packerCommand = packerCommandFactory.buildPackerCommand(baseCommand, parameterMap, "")
+
+    then:
+    packerCommand == expectedPackerCommand
+
+    where:
+    something | baseCommand | expectedPackerCommand
+    "sudo"    | "sudo"      | ["sudo", "packer", "build", "-color=false", "-var", "something=sudo"]
+    "null"    | null        | ["packer", "build", "-color=false", "-var", "something=null"]
+    "empty"   | ""          | ["packer", "build", "-color=false", "-var", "something=empty"]
+  }
+}
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -39,6 +39,20 @@ executionStatusToBakeResults:
 
 defaultCloudProviderType: aws
 
+# When the bakery is configured to use a templateFile in this list,
+# /usr/bin/packer will be run as root using 'sudo'.
+# By default, spinnaker does not have sudo permissions so these scripts will
+# fail.
+# In order to give sudo permissions, create and add the following line (without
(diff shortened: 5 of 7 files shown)


gardleopard:
Test with packer running a real chroot build:
```
amazon-chroot output will be in this color.

==> amazon-chroot: Prevalidating AMI Name...
==> amazon-chroot: Gathering information about this EC2 instance...
==> amazon-chroot: Inspecting the source AMI...
==> amazon-chroot: Checking the root device on source AMI...
==> amazon-chroot: Creating the root volume...
==> amazon-chroot: Attaching the root volume to /dev/sdf
==> amazon-chroot: Mounting the root device...
==> amazon-chroot: Mounting additional paths within the chroot...
    amazon-chroot: Mounting: /proc
    amazon-chroot: Mounting: /sys
    amazon-chroot: Mounting: /dev
    amazon-chroot: Mounting: /dev/pts
    amazon-chroot: Mounting: /proc/sys/fs/binfmt_misc
==> amazon-chroot: Copying files from host to chroot...
    amazon-chroot: /etc/resolv.conf
==> amazon-chroot: Provisioning with shell script: /opt/rosco/config/packer/install_packages.sh
    amazon-chroot: repository=
    amazon-chroot: package_type=deb
    amazon-chroot: packages=vim
    amazon-chroot: upgrade=
    amazon-chroot: creating /usr/sbin/policy-rc.d to prevent services from being started
    amazon-chroot: Ign http://archive.ubuntu.com trusty InRelease
    amazon-chroot: Get:1 http://archive.ubuntu.com trusty-updates InRelease [65.9 kB]
    amazon-chroot: Hit http://archive.ubuntu.com trusty Release.gpg
    amazon-chroot: Hit http://archive.ubuntu.com trusty Release
    amazon-chroot: Get:2 http://security.ubuntu.com trusty-security InRelease [65.9 kB]
    amazon-chroot: Get:3 http://archive.ubuntu.com trusty-updates/main amd64 Packages [730 kB]
    amazon-chroot: Get:4 http://archive.ubuntu.com trusty-updates/restricted amd64 Packages [15.9 kB]
    amazon-chroot: Get:5 http://archive.ubuntu.com trusty-updates/universe amd64 Packages [339 kB]
    amazon-chroot: Get:6 http://archive.ubuntu.com trusty-updates/multiverse amd64 Packages [13.2 kB]
    amazon-chroot: Get:7 http://archive.ubuntu.com trusty-updates/main Translation-en [367 kB]
    amazon-chroot: Get:8 http://archive.ubuntu.com trusty-updates/multiverse Translation-en [7227 B]
    amazon-chroot: Get:9 http://archive.ubuntu.com trusty-updates/restricted Translation-en [3699 B]
    amazon-chroot: Get:10 http://archive.ubuntu.com trusty-updates/universe Translation-en [182 kB]
    amazon-chroot: Get:11 http://security.ubuntu.com trusty-security/main amd64 Packages [444 kB]
    amazon-chroot: Hit http://archive.ubuntu.com trusty/main amd64 Packages
    amazon-chroot: Hit http://archive.ubuntu.com trusty/restricted amd64 Packages
    amazon-chroot: Hit http://archive.ubuntu.com trusty/universe amd64 Packages
    amazon-chroot: Hit http://archive.ubuntu.com trusty/multiverse amd64 Packages
    amazon-chroot: Hit http://archive.ubuntu.com trusty/main Translation-en
    amazon-chroot: Hit http://archive.ubuntu.com trusty/multiverse Translation-en
    amazon-chroot: Hit http://archive.ubuntu.com trusty/restricted Translation-en
    amazon-chroot: Hit http://archive.ubuntu.com trusty/universe Translation-en
    amazon-chroot: Get:12 http://security.ubuntu.com trusty-security/restricted amd64 Packages [13.0 kB]
    amazon-chroot: Get:13 http://security.ubuntu.com trusty-security/universe amd64 Packages [125 kB]
    amazon-chroot: Get:14 http://security.ubuntu.com trusty-security/multiverse amd64 Packages [4991 B]
    amazon-chroot: Get:15 http://security.ubuntu.com trusty-security/main Translation-en [242 kB]
    amazon-chroot: Get:16 http://security.ubuntu.com trusty-security/multiverse Translation-en [2570 B]
    amazon-chroot: Get:17 http://security.ubuntu.com trusty-security/restricted Translation-en [3206 B]
    amazon-chroot: Get:18 http://security.ubuntu.com trusty-security/universe Translation-en [73.8 kB]
    amazon-chroot: Fetched 2699 kB in 8s (335 kB/s)
    amazon-chroot: Reading package lists...
    amazon-chroot: Reading package lists...
    amazon-chroot: Building dependency tree...
    amazon-chroot: Reading state information...
    amazon-chroot: vim is already the newest version.
    amazon-chroot: 0 upgraded, 0 newly installed, 0 to remove and 160 not upgraded.
    amazon-chroot: removing /usr/sbin/policy-rc.d
==> amazon-chroot: Unmounting the root device...
==> amazon-chroot: Detaching EBS volume...
==> amazon-chroot: Creating snapshot...
    amazon-chroot: Snapshot ID: snap-f177fff5
==> amazon-chroot: Registering the AMI...
==> amazon-chroot: AMI: ami-154acf66
==> amazon-chroot: Waiting for AMI to become ready...
==> amazon-chroot: Adding tags to AMI (ami-154acf66)...
    amazon-chroot: Adding tag: "appversion": ""
    amazon-chroot: Adding tag: "build_host": ""
==> amazon-chroot: Tagging snapshot: snap-f177fff5
==> amazon-chroot: Deleting the created EBS volume...
Build 'amazon-chroot' finished.

==> Builds finished. The artifacts of successful builds are:
--> amazon-chroot: AMIs were created:

eu-west-1: ami-154acf66
```

gardleopard:
Debug output when running ebs builder:
```
Running mock
TERM=linux
SPINNAKER_AWS_ENABLED=false
SPINNAKER_GOOGLE_ENABLED=false
SPINNAKER_GOOGLE_DEFAULT_REGION=us-central1
NLSPATH=/usr/dt/lib/nls/msg/%L/%N.cat
PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin
PWD=/
XFILESEARCHPATH=/usr/dt/app-defaults/%L/Dt
SPINNAKER_GOOGLE_DEFAULT_ZONE=us-central1-f
SPINNAKER_GOOGLE_PROJECT_ID=
HOME=/home/spinnaker
SHLVL=1
UPSTART_INSTANCE=
UPSTART_JOB=rush
SPINNAKER_AWS_DEFAULT_REGION=us-west-2
_=/usr/bin/env
sleep
done sleeping
i am
spinnaker
sleep
```

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
```diff
@@ -20,8 +20,8 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
 
   @Override
   List<String> buildPackerCommand(String baseCommand, Map<String, String> parameterMap, String absoluteTemplateFilePath) {
-    def packerCommand = ["packer", "build", "-color=false"]
-
+    def packerCommand = [baseCommand, "packer", "build", "-color=false"]
+    packerCommand.removeAll([""])
```
duftler: Is the `removeAll([""])` needed here? If `CloudProviderBakeHandler.getBaseCommand()` already returned `""`, I would have expected this to work ok without removing the empty string item.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -132,6 +132,11 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
   }
 
   @Override
+  String getBaseCommand() {
+    return awsBakeryDefaults.templateFile == "aws-chroot.json" ? "sudo" : ""
```
duftler: Would it make more sense to add some config to rosco.yml identifying a list of templates that need to be run as `root`? If we hardcode just this template here, it make make it difficult to customize later.

duftler: I think you also want to add an entry for `aws-chroot.json` just after the ebs builder here: https://github.com/spinnaker/rosco/blob/master/rosco-web/rosco-web.gradle#L94

Will make it easier for users to safely upgrade the package.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
```diff
@@ -0,0 +1,44 @@
+package com.netflix.spinnaker.rosco.providers.util
+
+import spock.lang.Shared
+import spock.lang.Specification
+
+class LocalJobFriendlyPackerCommandFactorySpec extends Specification {
+
+    @Shared
+    LocalJobFriendlyPackerCommandFactory packerCommandFactory = new LocalJobFriendlyPackerCommandFactory()
+
```
duftler: Can you please format this class like the other test classes? (2-space indentation, `setup/when/then` content indented, trim some of the extraneous blank lines)

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
duftler: Come to think of it, if you go the config route this bit of logic can move to the base `getBaseCommand()`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
duftler: The extraneous `""` likely had no ill-effect because the tokenized command just gets space-joined here: https://github.com/spinnaker/rush/blob/master/src/main/groovy/com/netflix/spinnaker/rush/local/scripts/ScriptExecutorLocal.groovy#L58

(I'm thinking maybe you spotted the extra `""` when writing your unit tests and it looked a bit strange. If you think it reads better to strip it out here, I'm ok leaving it as you have it. Maybe also handle null here and in the unit tests, just in case any future added provider passes in null?)

gardleopard:
@duftler Thanks for reviewing, I will look into all this in some hours and give it another try.

ewiseblatt:
I still dont understand why you need root (or why this hasnt come up before). If it turns out you do, could you do something to restrict the process scope? You only need it in a job spawned by the bakery but the "spinnaker" user spans lots of services with a much wider surface area.

gardleopard:
@ewiseblatt I agree that it feels bad that rush can start jobs as root. Would it be a solution for this to run the rush service as a separate user? From what I've thought of this so far the most scary service to have root capabilities is rush and we are giving rush this capability. I see that there has been done work in packer to support running chroot without being root (https://github.com/mitchellh/packer/issues/430). As far as I can understand the scariest part here is to add the spinnaker user to sudoers for this, not that the packer command runs as root or just switches to root at will. 

The profit from using chroot instead  of ebs builder is huge. So for aws users this is a key feature to get fast pipelines imho.

ewiseblatt:
I dont know much about security other than to be fearful of this type of
thing. I've made some inquiries offline hoping to get some guidance from
someone knowledgable.

If we are giving/requiring root privileges, especially silently, I'd want
to restrict it as close to where actually needed as possible and not taint
the whole system or deployments that dont need that capability. Especially
since we're only talking about 1/8th of the services (that dont need to be
colocated), and not all deployments at that.

The discussion in your link was closed 2.5 years ago. Does that mean it was
resolved? I didnt read the details.

gardleopard:
Thanks for investigating @ewiseblatt.

Regarding the issue I linked, I think it is fixed from packers point of view. From spinnaker/rush point of view. I think it still needs the sudo permission to run, but it can start as a regular user and use root only for the steps in packer that requires root. So from how I understand, this is not a solution to the issue that the spinnaker user can become root.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
```diff
@@ -32,6 +31,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
     }
 
     packerCommand << absoluteTemplateFilePath
+    packerCommand.removeAll([null])
```
gardleopard: @duftler I did this to handle the null in the baseCommand and also for the rest of the packerCommand.

gardleopard:
@duftler I have changed the logic for running with sudo now. Its now a property in rosco.yml where you can set `runAsRoot: true` in you bakeryDefaults.

ewiseblatt:
One suggestion I received was to restrict the sudo permission to only the
commands you need to run. Another comment was that if the tool were run in
its own container then it might not matter that it could sudo to root.

I'm not that familiar with rush, but thought it used docker containers. If
that is the case, might it be viable to have this particular activity run
within an isolated container? I dont know if that makes sense or not (or
how viable it is). I'm just thinking out loud here.

Minimally, you should be able to restrict your sudo to the commands you
need as an initial step forward.

Maybe I'll get more response to my inquiry on Monday.

- eric

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
gardleopard: I did a test of the join function in rush. This test is successful, I imagine null values to rush can break stuff:
```
  void 'test the string'() {
    when:
      List<String> listOfStrings = [ "test ", "", "empty", "and", null, "null", "strings" ]

    then:
      String output = listOfStrings.join(" ")

    expect:
      output == "test   empty and null null strings"

  }
```

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -120,7 +120,7 @@ abstract class CloudProviderBakeHandler {
    * Returns the command that should be prepended to the shell command passed to the container.
    */
   String getBaseCommand() {
-    return ""
+    return getBakeryDefaults().runAsRoot ? "sudo" : ""
```
duftler: Would prefer to have a precise list of templates specified in rosco.yml that require root. Shortest path I can think of looks like:

1) Add this to `RoscoConfiguration`:
```
  // Allows @Value annotation to tokenize a list of strings.
  @Bean
  ConversionService conversionService() {
    new DefaultConversionService()
  }
```

2) Add this to `rosco.yml`:
```
templatesNeedingRoot: aws-chroot.json
```

3) Add this to `CloudProviderBakeHandler`:
```
  @Value('${templatesNeedingRoot:}')
  List<String> templatesNeedingRoot
```

4) Change `getBaseCommand()` to:
```
String getBaseCommand(String templateFileName) {
  return templatesNeedingRoot.contains(templateFileName) ? "sudo" : ""
}
```

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+    when:
+      def packerCommand = packerCommandFactory.buildPackerCommand("", parameterMap, templateFile)
+
+    then:
+      packerCommand == ["", "packer", "build", "-color=false", "-var", "something=slow", "aws-ebs.json"]
+  }
+
+  void "suports null baseCommand"() {
+    setup:
+    def parameterMap = [
```
duftler: Indentation is off in this test case.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
duftler: Makes sense. And the tests look good.

duftler:
@ewiseblatt I added a comment to the linked PR in spinnaker/spinnaker with the config tailored to just allow `sudo` for `packer`.
Regarding running these commands in containers, that's how it worked just prior to the oss release. We moved away from doing that due (primarily) to the friction involved in dealing with the credentials for each container image, and the need to operate yet another hosted system.
Seems relevant to mention here that spinnaker is being used to provision new vm's, that those vm's can have whatever scope the caller requests, that the vm's allow startup scripts, and that those startup scripts can execute arbitrary commands as root.
I do agree we should keep the sudo scope as narrow as possible, and also only allow it at all if the requested packer template strictly requires it (my final comments above relate to this restriction).

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
gardleopard: @duftler fixed in latest commit.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
gardleopard: @duftler this is fixed in latest commit.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -120,7 +123,11 @@ abstract class CloudProviderBakeHandler {
    * Returns the command that should be prepended to the shell command passed to the container.
    */
   String getBaseCommand() {
-    return getBakeryDefaults().runAsRoot ? "sudo" : ""
+    if(templatesNeedingRoot) {
```
duftler: Please insert a space before the `(`.

rosco-web/config/rosco.yml
```diff
@@ -39,12 +39,13 @@ executionStatusToBakeResults:
 
 defaultCloudProviderType: aws
 
+templatesNeedingRoot: [ aws-chroot.json ]
```
duftler: This value needs to be specified without the enclosing brackets:

`templatesNeedingRoot: aws-chroot.json`

It is not a proper yaml list, just a string representing a list of comma-separated values that the spring conversion stuff will tokenize.

duftler:
This LGTM.
Please squash all commits after addressing the 2 remaining comments.

Then we just need to wait until the companion PR is resolved (hopefully tomorrow).

duftler:
Thanks for putting this together. It's really a nice improvement.

gardleopard:
@duftler you are welcome, thanks for all the guidance! Last to fixes is done and also commits are squashed.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
```diff
@@ -32,6 +32,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
 
     packerCommand << absoluteTemplateFilePath
     packerCommand.removeAll([null])
+    packerCommand.removeAll("")
```
gardleopard: @duftler I just did a test of this on a running instance, this is critical for rush to work. 
Base command was never used before my pull request... https://github.com/gardleopard/rosco/blob/run_aws_chroot_with_sudo/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy#L23

duftler: Ok, then this should probably be just one line:
`packerCommand.removeAll([null, ""])`

gardleopard: ok, fixed and squashed:)

rosco-web/config/rosco.yml
```diff
@@ -41,6 +41,12 @@ defaultCloudProviderType: aws
```
ewiseblatt: Move the comment below up here before the attribute and reword it.

# When the bakery is configured to use a templateFile in this list, /usr/bin/packer will be run as root using 'sudo'.
# By default, spinnaker does not have sudo permissions so these scripts will fail.
# In order to give sudo permissions, create and add the following line (without leading '#') to /etc/sudoers.d/spinnaker
# spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer
#
# WARNING: Giving sudo access for spinnaker to execute packer may create an opportunity for malicious actors
# to take control of your machine and data it has access to.

rosco-web/config/rosco.yml
```diff
@@ -39,14 +39,16 @@ executionStatusToBakeResults:
...
-templatesNeedingRoot: aws-chroot.json
-
-# An entry in /etc/sudoers.d is needed to grant spinnaker sudo rights for packer.
-# This is a security risk that can be exploited.
-# file: /etc/sudoers.d/spinnaker
-# # User rules for spinnaker
+# When the bakery is configured to use a templateFile in this list, /usr/bin/packer will be run as root using 'sudo'.
+# By default, spinnaker does not have sudo permissions so these scripts will fail.
+# In order to give sudo permissions, create and add the following line (without leading '#') to /etc/sudoers.d/spinnaker
 # spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer
```
ewiseblatt: Could you add a # here so that it looks like a single comment, and also format it so it fits in 80 columns?
Then squash your commit if it wasnt already so I can merge it.

gardleopard: @ewiseblatt fixed comment column length and squashed commits. Ready for merge!

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -354,6 +358,71 @@ class AWSBakeHandlerSpec extends Specification {
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, "$configDir/$awsBakeryDefaults.templateFile")
   }
 
+  void 'produces packer command with all required parameters for amzn, with sudo using runAsRoot:true'() {
+    setup:
+      def awsBakeryDefaultsJson = [
+        awsAccessKey: "FOO",
```
ewiseblatt: we removed the aws keys from the rosco.bakeryDefaults last week
https://github.com/spinnaker/spinnaker/commit/8a2da11c9d29e3a879a068c197568c108b9e5f88

ewiseblatt: I see the other tests below use these too. Perhaps they should be removed from those as well. It probably isnt wrong, but is misleading. I think. I assume these arent injected anywhere at runtime.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -354,6 +358,71 @@ class AWSBakeHandlerSpec extends Specification {
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, "$configDir/$awsBakeryDefaults.templateFile")
   }
 
+  void 'produces packer command with all required parameters for amzn, with sudo using runAsRoot:true'() {
+    setup:
```
ewiseblatt: It might not be possible to do, but would be nice if the variations of producing packer commands could be factored out into a common function that has the boiler plate and templates so each specialized test just gives it how it is different than the others. That would clarify what the differences are and probably be more maintainable.

I've suggested this type of thing before and sometimes it turns out not to be possible. Dont kill yourself trying to do it, but see if it is reasonable to do.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
```diff
@@ -0,0 +1,52 @@
...
+  void "baseCommand is included"() {
+    setup:
+      def parameterMap = [
+          something:   "fast"
+      ]
+      String templateFile = "aws-chroot.json"
+
+    when:
+      def packerCommand = packerCommandFactory.buildPackerCommand("sudo", parameterMap, templateFile)
+
```
ewiseblatt: templateFile isnt considered when building the command, right?
So it seems you could have one function that takes a base command argument and expected prefix (list). Then your 3 tests call this function with ("sudo", ["sudo"]), (null, []) and ("", []). Not only is it simpler and clearer, it emphasizes that only the baseCommand is relevant, not the templateFile.

I havent written spock tests in a while so dont know if this makes sense or a good idea. @duftler may disagree.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
gardleopard: I fixed this with a new commit.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
ewiseblatt: On second thought, I am wondering if you need this test at all.
If you can show in other tests that you are calling baseCommand then you just need to add tests that show baseCommand return sudo when you want it to and not when you dont. Then the LocalJobFriendlyPackerCommandFactorySpec below shows that you are constructing the packer command correct. That gives you a more concise set of tests and eliminates a lot of the noise here with all the extra parameter decoration and tedious mocking.

1) you get sudo exactly when you want
2) you build command lines properly with and without sudo
3) your flow includes the calls to determine when you want sudo and will pass that decision into building a command (and uses all the other configuration stuff).

What do you think?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactorySpec.groovy
gardleopard: I have solved this in the latest commit

gardleopard:
@ewiseblatt I have done all the changes you requested and squashed the commits.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
gardleopard: The CloudProviderBakeHandler is an abstract class. I think it must be extended to be tested, so I choose to use AWSBakeHandler to test it. I could probably have done this differently, but I don't know how a better solution would be. Imho the test adds value because it is testing the getBaseCommand in the abstract class.

PR #81, opened 2016-03-25
Do not consider ami_suffix when composing the bake key.

duftler:


duftler:
@gardleopard please review.
@tomaslin fyi

duftler:
Since orca injects a timestamp in the place of a missing `amiSuffix` attribute (https://github.com/spinnaker/orca/blob/99639d35ec062ff622320f7cb774da1220a03fc1/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/pipeline/BakeStage.groovy#L86-L88), including `ami_suffix` in the bake key prevented the uniqueness check from ever identifying duplicates.

gardleopard:
Looks good to me.

PR #82, opened 2016-04-01
Change naming-convention for rpms is from name-version-release-arch t…

duftler:
…o name-version-release.arch.


duftler:
@diananana and @gardleopard please review.

PR #85, opened 2016-04-07
Force yum to preserve package installation order

kalymero:
This behavior is consitent with the debian one. Also yum correctly fail (also aborting the build)
if the package he's trying to install cannot be installed


duftler:
LGTM.

PR #86, opened 2016-04-10
Add support for us-west-1. Comment that images are placeholders.

dmourati:
Following up on a comment from:

https://github.com/kenzanlabs/spinnaker-terraform/blob/master/docs/AWS.md

This PR updates AWS AMI's to the latest. It adds a comment link to the source for AMI info:

https://cloud-images.ubuntu.com/locator/ec2/

Finally, it adds support for the us-west-1 region mentioned in the spinnaker-terraform docs above.

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -64,6 +64,8 @@ aws:
     # TODO(duftler): Support additional store types beyond ebs.
     # TODO(duftler): Support additional operating systems.
     # TODO(duftler): Feels like just about everything here could be (optionally) specified per bake.
+    # AMIs sourced from: https://cloud-images.ubuntu.com/locator/ec2/
+    # Images should be considered placeholders
     - baseImage:
         id: ubuntu
         shortDescription: v12.04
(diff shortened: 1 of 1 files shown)


duftler:
We had always thought of the default configured ami's as placeholders. This does raise a couple of questions though:

1) How do we know if the new ami's are the 'right' ones?
2) How often do we need to update this config?

My impression was that the default configured ami's would be enough for people to experiment with baking (and work through the various Spinnaker tutorials), but that they would then go and hand-pick a curated set of ami's (likely org-specific images with some base software layer installed) for their real deployments.

I'm ok updating to a later set of ami's and adding defaults for the additional regions, but would like to avoid an ongoing maintenance headache w.r.t. having to keep updating that list.

dmourati:
I agree and just need west 1 to do the real deployments. Perhaps a Jenkins job to poll the official Amis writing to a conf override local dot file.

duftler:
Ok, so for now do you want to update the pending PR to add just the defaults for the two new regions (and maybe a comment that these are to be considered placeholders)?

dmourati:
Updated per conversaion @duftler .

duftler:
LGTM.

PR #88, opened 2016-04-29
Parse a FQPN at any position in the package list

kalymero:
This patch is a step forward in correct handling multiple packages. With this patch if a FullQualifiedPackageName get passed in the
package list it get parsed no matter the position (before it only happened for the first one).
Also, the first package keep its importance of potential determining the ami-name or the Appversion if correctly parsed

This is related to: https://github.com/spinnaker/orca/pull/800 : packages can come down in any order

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
@@ -39,14 +39,13 @@ public class DefaultImageNameFactory implements ImageNameFactory {
     def timestamp = clock.millis()
 
     List<String> packageNameList = bakeRequest?.package_name?.tokenize(" ")
-    PackageNameConverter.OsPackageName osPackageName
     String appVersionStr
-    AppVersion appVersion
-    def baseImagePackageName
+    String baseImagePackageName
 
     if (packageNameList) {
       packageNameList.eachWithIndex { packageName, index ->
 
+        PackageNameConverter.OsPackageName osPackageName
         osPackageName = PackageNameConverter.buildOsPackageName(selectedOptions.baseImage.packageType, packageName)
 
         if (osPackageName?.name) {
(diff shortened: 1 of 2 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
```diff
@@ -39,35 +39,43 @@ public class DefaultImageNameFactory implements ImageNameFactory {
     def timestamp = clock.millis()
 
     List<String> packageNameList = bakeRequest?.package_name?.tokenize(" ")
-    def firstPackageName
     PackageNameConverter.OsPackageName osPackageName
     String appVersionStr
     AppVersion appVersion
+    def baseImagePackageName
```
duftler: Can use `String` here since the neighbors do.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
```diff
@@ -39,35 +39,43 @@ public class DefaultImageNameFactory implements ImageNameFactory {
...
 
-        // If appVersionStr could not be parsed to create AppVersion, clear it.
-        appVersionStr = null
-      }
+          // If a version/release was specified, we need to include that when installing the package.
+          if (osPackageName?.version && osPackageName?.release) {
+            packageNameList[index] += "$selectedOptions.baseImage.packageType.versionDelimiter$osPackageName.version-$osPackageName.release"
+          }
+
+          // First package is special, its name and its attributes may be used to build the AMI name and add additional
```
duftler: s/AMI/image

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
```diff
@@ -39,35 +39,43 @@ public class DefaultImageNameFactory implements ImageNameFactory {
...
+          // attributes (eg: AppVersion)
```
duftler: s/AppVersion/appversion & build_host

duftler: Please add a terminating period to the comment.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
```diff
@@ -39,35 +39,43 @@ public class DefaultImageNameFactory implements ImageNameFactory {
...
+          if (index == 0) {
+            // We need to replace the original fully-qualified package name for with the unqualified
```
duftler: s/for with/with

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/DefaultImageNameFactory.groovy
```diff
@@ -39,35 +39,43 @@ public class DefaultImageNameFactory implements ImageNameFactory {
...
     if (packageNameList) {
-      // For now, we only take into account the first package name when generating the appversion tag.
-      firstPackageName = packageNameList[0]
+      packageNameList.eachWithIndex { packageName, index ->
 
-      // Passing in firstPackageName here to avoid tokenizing the package name list twice.
-      osPackageName = PackageNameConverter.buildOsPackageName(selectedOptions.baseImage.packageType, firstPackageName)
-      appVersionStr = PackageNameConverter.buildAppVersionStr(bakeRequest, osPackageName)
-      appVersion = AppVersion.parseName(appVersionStr)
+        osPackageName = PackageNameConverter.buildOsPackageName(selectedOptions.baseImage.packageType, packageName)
```
duftler: Since `osPackageName` can get re-assigned multiple times now, can it be defined here instead of the outer scope? Same for `appVersion`?

duftler: Actually, maybe we shouldn't even capture the response from `AppVersion.parseName(appVersionStr)` in a variable at all?

duftler:
LGTM after addressing comments.

duftler:
LGTM.

PR #89, opened 2016-05-05
Improve the unicity of the bake key also taking into account multiple accounts

kalymero:
This PR address the following issue: Given two bakes with the same ami-name but different installed package(s), the bakeKey should be different, otherwise a rebake is required to bake it again (also the ami-suffix is not part of the bakeKey).
This is very evident when baking from the Bake stage of a pipeline. The ami-name is constant at this point, while the installed packages version is different for each build

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
@@ -79,7 +79,7 @@ abstract class CloudProviderBakeHandler {
         bakeKey += ":$ami_name"
       }
 
-      String packages = package_name.tokenize().join('-')
+      String packages = package_name.tokenize().join('|')
       bakeKey += ":$packages"
 
       def providerSpecificBakeKeyComponent = produceProviderSpecificBakeKeyComponent(region, bakeRequest)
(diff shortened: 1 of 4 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -77,10 +77,11 @@ abstract class CloudProviderBakeHandler {
 
       if (ami_name) {
         bakeKey += ":$ami_name"
-      } else {
-        bakeKey += ":$package_name"
       }
 
+      String packages = package_name.tokenize().join('-')
```
duftler: The only issue I see here is that there is a potential naming-collision. These two bake requests would result in the same token for the `packages` segment of the bake key:
`package_name == 'a-b c'`
`package_name == 'a b-c'`

Can avoid this (or at least greatly reduce the likelihood of it occurring) by choosing as the delimiter a character that is unlikely to appear in a package name.

duftler:
LGTM after addressing one comment.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
kalymero: right, I'll use a | for it

PR #91, opened 2016-05-11
Add support for password on a redis connection

JargoonPard:
Needed to add a timeout value since the JedisPool constructor requires one if you want to include a password.  This change supports using the Azure hosted Redis service.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/config/JedisConfig.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/config/JedisConfig.groovy
@@ -29,13 +29,14 @@ import redis.clients.jedis.JedisPool
 class JedisConfig {
 
   @Bean
-  JedisPool jedisPool(@Value('${redis.connection:redis://localhost:6379}') String connection) {
+  JedisPool jedisPool(@Value('${redis.connection:redis://localhost:6379}') String connection,
+                      @Value('${redis.timeout:2000}') int timeout) {
     RedisConnectionInfo connectionInfo = RedisConnectionInfo.parseConnectionUri(connection)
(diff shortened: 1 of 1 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/config/JedisConfig.groovy
```diff
@@ -29,13 +29,14 @@ import redis.clients.jedis.JedisPool
...
-  JedisPool jedisPool(@Value('${redis.connection:redis://localhost:6379}') String connection) {
+  JedisPool jedisPool(@Value('${redis.connection:redis://localhost:6379}') String connection,
+                      @Value('${redis.timeout:2000}') int timeout) {
     RedisConnectionInfo connectionInfo = RedisConnectionInfo.parseConnectionUri(connection)
     GenericObjectPoolConfig poolConfig = new GenericObjectPoolConfig()
     poolConfig.setMaxTotal(100)
     poolConfig.setMinIdle(25)
     poolConfig.setMaxIdle(100)
-    new JedisPool(poolConfig, connectionInfo.host, connectionInfo.port)
+    new JedisPool(poolConfig, connectionInfo.host, connectionInfo.port, timeout, connectionInfo.password)
```
duftler: What is the behavior of the `JedisPool` constructor if no password is configured?

JargoonPard: It behaves fine and works as expected.  This change aligns the constructor usage closer with how it is called in clouddriver. The only difference is in clouddriver a null is supplied rather than an empty string if no password is supplied. (line 64 of com.netflix.spinnaker.clouddriver.core.RedisConfig.groovy)

duftler:
LGTM.

PR #92, opened 2016-05-13
Make it work for eu-central-1 and eu-west-1

schmidtwastl:


duftler:
Thanks for the contribution.

There is a comment at the top of that section: `# AMIs sourced from: https://cloud-images.ubuntu.com/locator/ec2/`

I took a quick look at that table and didn't spot the two newly-included ami's (it's possible I misread the table, as I'm not an expert on selecting ami's). Is there a reason not to select the new ami id's from that table?

schmidtwastl:
You are right, it´s not in the table. However, the other AMIs cannot be found in the table anymore as well. I´ve just checked it for the us-east-1 and us-west-1 AMIs. However, they can still be found in the AWS Console if you search for them under "Community AMIs". So they are still available for an instance launch. That´s why it still works.

My way of selecting the Ubuntu 14.04 AMI is 
1. go to the AWS console 
2. make sure you are in the correct region
3. Start the ec2-launch-wizard and copy the AMI-id from the quick-start screen

Advantage: This AMI-id doesn´t change regularly. 

If you want, I can replace them with the respective AMI-ids from the table.

For production use I would use my own AMIs anyway. However, it would be nice, if I could specify the Base AMI-id in the Browser instead of changing it in the code. (Perhaps it´s already possible. I was just trying out the tutorial for now)

ami

duftler:
SGTM (the logic to find appropriate ami id's).

duftler:
Btw, to specify an explicit source image directly in the bake configuration, you can do this:
image

That relies on the mechanism for adding additional key/value pairs to the packer call or for overriding pre-defined user variables like this one: https://github.com/spinnaker/rosco/blob/master/rosco-web/config/packer/aws-ebs.json#L10

duftler:
That configured bake stage results in this expected error upon execution:
```
amazon-ebs output will be in this color.

==> amazon-ebs: Prevalidating AMI Name...
==> amazon-ebs: Inspecting the source AMI...
==> amazon-ebs: Error querying AMI: InvalidAMIID.Malformed: Invalid id: "some-ami" (expecting "ami-...")
==> amazon-ebs: 	status code: 400, request id: []
Build 'amazon-ebs' errored: Error querying AMI: InvalidAMIID.Malformed: Invalid id: "some-ami" (expecting "ami-...")
	status code: 400, request id: []

==> Some builds didn't complete successfully and had errors:
--> amazon-ebs: Error querying AMI: InvalidAMIID.Malformed: Invalid id: "some-ami" (expecting "ami-...")
	status code: 400, request id: []

==> Builds finished but no artifacts were created.
```

duftler:
Just realized that the suggestion I made for specifying a base image is not the most direct way. There exists a field for this purpose: 
image

schmidtwastl:
Nice! This was exactly what I was looking for!
Thanks for the first suggestion as well. I just tried it out as well. Very happy to see how much I can specify directly from the UI!

PR #94, opened 2016-05-15
Remove config files with possible credentials from baked AMIs

amanya:
Remove configuration files from either debian and yum repos after installing packages. We are doing this because we put basic auth credentials in repo urls.


duftler:
LGTM. Thanks for the contribution.

PR #98, opened 2016-05-19
updating authors

drmaas:


duftler:
Not sure why, but there doesn't seem to be an associated travis build for this pr (and it won't let me merge a pr without a successful travis build). Feel free to just add this change into the open README pr.

PR #99, opened 2016-05-20
Add Schibsted to Authors.

duftler:


duftler:
@gardleopard @kalymero fyi

PR #105, opened 2016-06-15
Refactor rosco to run jobs directly instead of relying on rush.

duftler:
Drop dependency on cassandra.


duftler:
@lwander or @ttomsu please review.
@gardleopard please review.

duftler:
The refactored rosco will respect pre-existing bakes in the same redis cluster (including in-flight bakes).

gardleopard:
Great @duftler! I have tested this in our dev environment and it works. Parallel building works and it does not interfere with old version of rosco/rush running bakes with the same redis backend.

Making the rush dependency and at the same time removing 257 lines of code is awesome! 
screen shot 2016-06-15 at 22 28 06 👍

build.gradle
```diff
@@ -20,7 +20,7 @@ allprojects {
   group = "com.netflix.spinnaker.rosco"
 
   spinnaker {
-    dependenciesVersion = "0.19.0"
+    dependenciesVersion = "0.40.0"
```
ttomsu: Yikes - that's a jump!

ttomsu:
LGTM

PR #107, opened 2016-06-20
Changed provision_deb so it can take multiple repos

dennismartensson:
Have change the provision_deb functions so it loops over $repository split on ;

This makes it possible to set more the one deb repo in debianRepository and have them all added. I did split the settings string on ; with to my knowledge is a non valid character for the urls and there by safe as a character to split on.

Please add comments.


duftler:
LGTM.

PR #108, opened 2016-06-21
Set HOME env when running from deb package.

duftler:
When I merged the job-runner logic from rush into rosco, I neglected to add the config to set HOME env when running rosco from an installed .deb package.
This omission left the aws builder for packer unable to resolve .aws/credentials: https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/shared_credentials_provider.go#L125


duftler:
@lwander please review.

lwander:
LGTM

PR #109, opened 2016-06-22
First pass at collecting metrics.

duftler:


duftler:
Can `curl localhost:8087/metrics` to see results.

duftler:
@ewiseblatt please review.

build.gradle
```diff
@@ -20,7 +20,7 @@ allprojects {
   group = "com.netflix.spinnaker.rosco"
 
   spinnaker {
-    dependenciesVersion = "0.40.0"
```
ewiseblatt: Does this depend on a new version of dependencies?

duftler: Yes. Released it here:
https://github.com/spinnaker/spinnaker-dependencies/tree/v0.41.0

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
```diff
@@ -161,14 +169,23 @@ class JobExecutorLocal implements JobExecutor {
...
+    // Terminate the process.
+    canceledJob?.watchdog?.destroyProcess()
 
     // The next polling interval will be unable to retrieve the job status and will mark the bake as canceled.
   }
+
+  @PostConstruct
+  void initializeMetrics() {
+    registry.gauge(registry.createId("bakes.local").withTag("active", "true"), jobIdToHandlerMap, new ToDoubleFunction<Map>() {
+
```
ewiseblatt: Why double when it is scalar?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -144,18 +154,43 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
...
+    Id bakesId = registry.createId('bakes')
 
     if (bakeStatus) {
       if (bakeStatus.state == BakeStatus.State.COMPLETED) {
         completeBake(statusId, bakeStatus.logsContent)
-      }
 
-      bakeStore.updateBakeStatus(bakeStatus)
+        bakeStore.updateBakeStatus(bakeStatus)
+
```
ewiseblatt: I think if you put a "Id bakesId" in the outer scope and determine the ID you want to update here, then you can make this if block just determine the id, then factor out the  bakeStore.updateBakeStatus and registry.timer(bakesId) after the if block in one place that all cases flow through.

duftler: Not all paths call `registry.timer()`.

duftler: I can probably only make the call if the id is actually set. Ok, let me try to rework it a bit.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
duftler: I used the double because that's what the `registry.gauge()` method expected. I couldn't find a non-double version.

duftler: Came out much nicer. PTAL.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
duftler: Came out much nicer. PTAL.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -120,6 +125,11 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
                             if (!cancellationSucceeded) {
                               bakeStore.removeFromIncompletes(roscoInstanceId, statusId)
                             }
+
+                            // This will have the most up-to-date timestamp.
```
ewiseblatt: I dont understand the control flow of how you get here. I'm wondering if this is double counted with other failures?

ewiseblatt:
LGTM, even if double counting. If you are double counting (e.g. timeouts are a subset of job failures), document that somewhere.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
duftler: This path specifically handles the case where the polling agent detects incomplete bakes that are not managed by _this_ rosco instance, and have not been updated beyond some threshold. If this instance is updating the bake, we will not end up here.

duftler: The incomplete bakes are scoped by rosco instance id, and we only take this branch if the managing instance is not _this_ instance.

duftler:
It shouldn't be double-counting. The path you called out relates to rosco instances other than this one, and timeouts on this instance are only counted via the 'cancellation' path in `BakePoller`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
ewiseblatt: oh, so this is really just an orphan due to a timeout and you are presuming that the managing rosco instance is negligent (e.g. died). I was assuming that we were the ones ophaning the task. A comment would be useful here.

duftler: Ok, will add a comment.

duftler: Added comment here: https://github.com/spinnaker/rosco/pull/110

PR #112, opened 2016-06-23
Removing the handling of the docker service from the DockerBakeHandler

brujoand:
The 'START_DOCKER_SERVICE_BASE_COMMAND' breaks for us when running docker bakes, so locally I removed it rather than trying to find a way to have it work, as I felt that it didn't belong under the responsibility of Rosco to ensure that the docker service was running. And as @duftler pointed out on the slack channel yesterday:

That is almost certainly a vestige of rush's prior incarnation of running bake jobs as containers themselves.


duftler:
Thanks for the contribution.
LGTM.

PR #118, opened 2016-07-06
Support the share_with and copy_to parameters as extended_attributes.

kalymero:
The packer part is quite ugly since it's not possible to pass a list on the command line as `var`.
Accounts and regions must me passed one by one

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
@@ -168,12 +168,29 @@ abstract class CloudProviderBakeHandler {
     parameterMap.configDir = configDir
 
     if (bakeRequest.extended_attributes) {
-      parameterMap << bakeRequest.extended_attributes
+      if (bakeRequest.extended_attributes.containsKey('share_with')) {
+        unrollParameters("share_with_", bakeRequest.extended_attributes.get('share_with'), parameterMap)
+      }
+
+      if (bakeRequest.extended_attributes.containsKey('copy_to')) {
+        unrollParameters("copy_to_", bakeRequest.extended_attributes.get('copy_to'), parameterMap)
+      }
+
+      List attributes = bakeRequest.extended_attributes.keySet().asList()
+      parameterMap << bakeRequest.extended_attributes.subMap(
+              attributes.findAll { !it.equals('share_with') && !it.equals('copy_to') })
     }
 
     def finalTemplateFileName = bakeRequest.template_file_name ?: templateFileName
 
     return packerCommandFactory.buildPackerCommand(baseCommand, parameterMap, "$configDir/$finalTemplateFileName")
+  }
+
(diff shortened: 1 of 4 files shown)


ajordens:
Needs a rebase @kalymero.

Can you give it a once over @duftler, I'm not a `rosco` user but in theory it seems sensible to me.

`ami_users` is a little less important (right?) given that Spinnaker will on-demand share to an account (_ie. it doesn't need to be shared at bake time_).

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -168,14 +168,33 @@ abstract class CloudProviderBakeHandler {
...
+      parameterMap << bakeRequest.extended_attributes.subMap(
+              attributes.findAll{ !it.equals('share_with') && !it.equals('copy_to') })
     }
 
     def finalTemplateFileName = bakeRequest.template_file_name ?: templateFileName
 
     return packerCommandFactory.buildPackerCommand(baseCommand, parameterMap, "$configDir/$finalTemplateFileName")
   }
 
+  private void unrollParameters(String prefix, String rolledParameter, Map parameterMap) {
```
duftler: Can drop the semi-colons in this method.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -168,14 +168,33 @@ abstract class CloudProviderBakeHandler {
...
+    List<String> values = rolledParameter.tokenize(",")
+    int i=1;
```
duftler: Would be more idiomatic to use `eachWithIndex` (http://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/Object.html#eachWithIndex(groovy.lang.Closure) here.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -859,4 +859,108 @@ class AWSBakeHandlerSpec extends Specification {
       bakeKey1 == "bake:aws:centos:kato|nflx-djangobase-enhanced_0.1-h12.170cdbd_all|mongodb:us-east-1:hvm:enhancedNWDisabled"
       bakeKey2 == bakeKey1
   }
+
```
duftler: Indentation in these new test methods doesn't match the neighbors.

duftler:
LGTM once all comments are addressed.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
kalymero: ohhh yes

kalymero:
it should be fine now

duftler:
LGTM.

PR #120, opened 2016-07-07
Add openstack metadata to image build

drmaas:
Also:

* filter openstack passwords in packer logs
* support v2 and v3 openstack packer templates

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
@@ -26,5 +26,16 @@ import groovy.transform.Immutable
 @CompileStatic
 class JobRequest {
   List<String> tokenizedCommand
+  List<String> maskedPackerParameters = []
+
+  List<String> getMaskedTokenizedCommand() {
+    return tokenizedCommand.collect { String masked ->
+      String key = masked.split('=').first()
+      if (key && key in maskedPackerParameters) {
+        masked = "$key=******".toString()
+      }
+      return masked
+    }
+  }
 }
 
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
@@ -185,4 +185,13 @@ abstract class CloudProviderBakeHandler {
     }
     return osVirtualizationSettings.baseImage
   }
+
+  /**
+   * This allows providers to specify Packer variables names whose values will be masked with '******'
+   * in logging output. Subclasses should override this if they need to mask sensitive data sent to Packer.
+   * @return
+   */
+  List<String> getMaskedPackerParameters() {
+    return []
+  }
 }
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -202,13 +202,13 @@ openstack:
   bakeryDefaults:
     identityEndpoint: ${OS_AUTH_URL:}
     domainName: Default
-    floatingIpPool: ext_vlan1472_net
+    floatingIpPool: ext_network
     securityGroups: default
-    tenantName: PRD-spinnakeropenstack
+    tenantName: tenant
     insecure: true
     username: ${OS_USERNAME}
     password: ${OS_PASSWORD}
-    templateFile: openstack.json
+    templateFile: openstack-v3.json
     baseImages:
     - baseImage:
         id: vivid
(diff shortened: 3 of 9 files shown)


drmaas:
@spinnaker/reviewers

rosco-web/config/rosco.yml
```diff
@@ -202,25 +202,25 @@ openstack:
   bakeryDefaults:
     identityEndpoint: ${OS_AUTH_URL:}
     domainName: Default
-    floatingIpPool: ext_vlan1472_net
+    floatingIpPool: ext_network
```
derekolk: Can we change these options use environment variables like endpoint?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
```diff
@@ -26,5 +27,17 @@ import groovy.transform.Immutable
 @CompileStatic
 class JobRequest {
   List<String> tokenizedCommand
+  List<String> packerParameterFilters = []
+
+  List<String> getFilteredTokenizedCommand() {
+    Lists.newArrayList(tokenizedCommand).collect {
```
duftler: What's the reason you need to use `newArrayList` if `tokenizedCommand` is already a `List`?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -185,4 +185,13 @@ abstract class CloudProviderBakeHandler {
     }
     return osVirtualizationSettings.baseImage
   }
+
+  /**
+   * packer variables names whose values will be replaced by '******'.
```
duftler: Please capitalize the start of the comment and add that this only affects logging output.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -185,4 +185,13 @@ abstract class CloudProviderBakeHandler {
...
+   * Subclasses should override this.
+   * @return
+   */
+  List<String> getPackerParameterFilters() {
+    []
```
duftler: Prefer explicit `return` here.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
duftler: Prefer explicit `return` here.

rosco-web/config/rosco.yml
drmaas: Why? The ones that currently use environment variables are tied to the environment variables that are provided via the openstack rc script that can be downloaded from the web console.

afaik you can do this in Spring boot for environment variables:

```
OPENSTACK_BAKERYDEFAULTS_FLOATINGIPPOOL=ext_network
```

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
```diff
@@ -27,16 +26,16 @@ import groovy.transform.Immutable
 @CompileStatic
 class JobRequest {
   List<String> tokenizedCommand
-  List<String> packerParameterFilters = []
+  List<String> maskedPackerParameters = []
 
-  List<String> getFilteredTokenizedCommand() {
-    Lists.newArrayList(tokenizedCommand).collect {
+  List<String> getMaskedTokenizedCommand() {
+    return tokenizedCommand.collect {
```
duftler: Might be cleaner with: `return tokenizedCommand.collect { String masked ->`

duftler:
LGTM as soon as final comment as addressed.
@derekolk Has your comment been addressed satisfactorily?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
```diff
@@ -26,5 +26,17 @@ import groovy.transform.Immutable
 @CompileStatic
 class JobRequest {
   List<String> tokenizedCommand
+  List<String> maskedPackerParameters = []
+
+  List<String> getMaskedTokenizedCommand() {
+    return tokenizedCommand.collect {
+      String masked = it
+      String key = it.split('=')[0]
```
duftler: Could also use `.first()` here instead of [0].

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
```diff
@@ -26,5 +26,17 @@ import groovy.transform.Immutable
...
+      if (key && maskedPackerParameters.contains(key)) {
```
duftler: Could use `in` operator instead of `contains()`, but I think it reads fine as-is.

rosco-web/config/rosco.yml
derekolk: I didn't know that these are upstream driven configuration ... Fine with leaving it as a generic default.

derekolk:
@duftler LGTM

duftler:
LGTM.

PR #121, opened 2016-07-08
Seperating the building of imageName, appVersionStr and packagesParameter

brujoand:
The rationale for this is that when baking docker images in Rosco the image names ends up quite convoluted, when the desired result is simply having the ami-name as image name, maybe with the possibility to add tags as appVersionStr. By splitting up the building of imageName, appVersionStr and packagesParameter I hope to make it easier to enable provider specific naming conventions. I looked at the possibility of just using 'extended_attributes', but to me this seems like something which will be needed by others also, so I went for the long road instead. 

This is my second stab at groovy so feel free to nitpick. Also, I'm on vacation, so I apologise if response isn't immediate.


lwander:
This might be to early to ask - but how do you plan on including the "organization" in the docker image name? Usually docker images follow the convention

```
<registry.domain>/<organization>/<image>:<tag>
```

Since "organization" is now a first class object in the [Docker Trigger configuration](https://github.com/spinnaker/deck/pull/2306#issuecomment-223418275), it would make sense to surface that somehow here.

brujoand:
Good question. My initial thought would be to expect an ami_name on the form of "myregistry.io/organization/repository" and then override buildImageName by parsing the ami_name. I guess this would be the easiest route, but increases probability of error. It also has to work with subsets of that ami_name etc. 
But this is very much colored by the fact that I hadn't noticed the change in the docker trigger configuration. What's your thoughts on this?

lwander:
I think ideally we would surface Registry (a pre-configured docker account in clouddriver), Organization and Image all as parameters to the build, which will likely be populated by dropdown. Tag will then be a function of package details/commit hash/date/etc... to ensure it's unique. This way it remains consistent with other references to Docker in Spinnaker, and will be easier to integrate with Kubernetes.

brujoand:
Sounds like a good approach to me. Do you see these new arguments passed to the build as explicit docker parameters or added to the extended attributes? 

While on the topic; the handling of docker tags has been puzzling me a bit. Some of those I've talked to wants to be able to use multiple tags pr image. But I guess that could be fitted in the extended attributes. Have you given that part any thought?

lwander:
I think where the parameters go is best left to @duftler (he's OOO so he might not respond for a bit).

When you mean multiple tags per image, do you mean the same (identical contents) image is tagged with several tags? I think that's a good practice in general and we should enable that.

brujoand:
Cool :thumbsup:
Yeah, same contents with multiple tags would be nice.

brujoand:
Merged with master, fixed conflict, squashed commits and pushed with force.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -140,7 +141,21 @@ abstract class CloudProviderBakeHandler {
...
+
+    List<String> packageNameList = bakeRequest?.package_name?.tokenize(" ")
+    def appVersionStr
+
+    def osPackages = packageNameList.collect { name ->
+      PackageNameConverter.buildOsPackageName(selectedOptions.baseImage.packageType, name)
+    }
+
+    if (!osPackages.empty) {
+     appVersionStr = imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.first())
```
duftler: Any reason not to make `imageNameFactory.buildAppVersionStr()` take the full list of packages, instead of extracting the first package here at the call-site, so it is more consistent with its siblings?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -66,6 +70,7 @@ class PackageNameConverter {
   // For example: nflx-djangobase-enhanced-0.1-h12.170cdbd.all
   public static OsPackageName parseRpmPackageName(String fullyQualifiedPackageName) {
     OsPackageName osPackageName = new OsPackageName()
+    String versionDelimiter = BakeRequest.PackageType.RPM.getVersionDelimiter()
```
duftler: Noticed that you refer to `versionDelimiter` up above against `DEB`, but used `getVersionDelimiter()` here for `RPM`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -82,6 +87,9 @@ class PackageNameConverter {
           release = parts.pop()
           version = parts.pop()
           name = parts.join("-")
+          qualifiedName = "${name}${versionDelimiter}${version}-${release}"
```
duftler: Since `qualifiedName` is composed the same way for both package types, maybe just make it a method on `OsPackageName`? It can still return `null` if `release` is missing.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
```diff
@@ -16,27 +16,73 @@
...
+    String arch = basePackage.arch ?: "all" // TODO This is defaulted to 'all' since all tests assume this. Skip?
+
+    [ baseName, arch, release, bakeRequest.base_os].findAll{it != "" }.join("-")
+  }
+
+  /**
+   * Returns packagesParameter; the updated list of packages to be used for overriding the passed package list
+   */
+  def buildPackagesParameter(List<PackageNameConverter.OsPackageName> osPackageNames) {
+    if (osPackageNames.empty) return ""
```
duftler: You should verify this, but I think is unnecessary.  `.collect` should return an empty list, which should produce an empty string when `.join` is applied to it.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -140,7 +141,21 @@ abstract class CloudProviderBakeHandler {
     def virtualizationSettings = findVirtualizationSettings(region, bakeRequest)
 
     BakeOptions.Selected selectedOptions = new BakeOptions.Selected(baseImage: findBaseImage(bakeRequest))
-    def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+
+    List<String> packageNameList = bakeRequest?.package_name?.tokenize(" ")
```
duftler: Since `packageNameList` can end up `null`, and the `.collect` down below will return an empty list when applied to it, the `osPackageNames.first()` just inside of `buildImageName()` will likely throw an exception when `package_name` is `null` or an empty string.

duftler: I didn't get to the unit tests yet, but we should probably have explicit tests for a `null` or empty string `package_name`.

duftler: `bakeRequest` can't really be `null` here, so we don't need the null safe navigation between it and `package_name`.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandlerSpec.groovy
```diff
@@ -159,6 +160,9 @@ class GCEBakeHandlerSpec extends Specification {
                                         base_os: "ubuntu",
                                         cloud_provider_type: BakeRequest.CloudProviderType.gce)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
+      def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
```
duftler: This pattern appears enough that it might make sense to extract it into a utility function for use in all the unit tests.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
```diff
@@ -16,27 +16,73 @@
...
+  }
+
+  /**
+   * Produces an imageName either from the BakeRequest.ami_name or the first package to be installed.
+   * This is to be used for naming the image being baked. Note that this function is not required to
+   * return the same image name on multiple invocations with the same bake request
+   */
+  def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
+    String timestamp = clock.millis()
+    def emptyBasePackage = new PackageNameConverter.OsPackageName()
```
duftler: This feels a little like abusing the `OsPackageName` model, especially since the `buildOsPackageName()` factory method throws an exception if it's not `DEB` or `RPM`. Maybe add an explicit type there for this use case?
If you decide you really want to keep it this way, you should be able to use: `def emptyBasePackage = new PackageNameConverter.OsPackageName(name: '', arch: 'all')`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
```diff
@@ -16,27 +16,73 @@
...
+      arch = "all"
+    }
+
+    PackageNameConverter.OsPackageName basePackage = osPackageNames.first() ?: emptyBasePackage
+
+    String baseName = bakeRequest.ami_name ?: basePackage.name
+    String release = bakeRequest.ami_suffix ?: timestamp
+    String arch = basePackage.arch ?: "all" // TODO This is defaulted to 'all' since all tests assume this. Skip?
+
+    [ baseName, arch, release, bakeRequest.base_os].findAll{it != "" }.join("-")
```
duftler: Removing the leading space before `baseName`, or adding a trailing space after `bakeRequest.base_os`, would increase readability. Same for the space surrounding `it != ""`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
```diff
@@ -16,27 +16,73 @@
...
+    def emptyBasePackage = new PackageNameConverter.OsPackageName()
+    emptyBasePackage.with {
+      name = ""
+      arch = "all"
+    }
+
+    PackageNameConverter.OsPackageName basePackage = osPackageNames.first() ?: emptyBasePackage
+
+    String baseName = bakeRequest.ami_name ?: basePackage.name
+    String release = bakeRequest.ami_suffix ?: timestamp
```
duftler: Might be more readable to have the order here match the order they are used in composing the image name down below.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
duftler: Since the test you're using here is `it != ""`, this leaves us open to concatenating `null` when one of the items in the list is explicitly `null`. While this looks impossible with the current path to this line, I think it would be more future-proof if the test were actually just `it`. Then the result would be false for both an empty string and a `null`.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -142,15 +155,20 @@ class DefaultImageNameFactorySpec extends Specification {
...
                                         base_os: "ubuntu")
 
+      def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
+        PackageNameConverter.parseDebPackageName(packageName)
+      }
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.head())
```
duftler: Using `head()` in some spots and `first()` in others may result in future readers searching around to see if they behave the same. We should probably just stick to one (I prefer `first()` but don't feel that strongly about it).

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -140,7 +141,21 @@ abstract class CloudProviderBakeHandler {
...
+    def appVersionStr
+
+    def osPackages = packageNameList.collect { name ->
```
duftler: Any reason not to call this `osPackageNames`, like the parameter names in the methods it is passed to?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -628,7 +672,7 @@ class AWSBakeHandlerSpec extends Specification {
         aws_target_ami: targetImageName,
         repository: DEBIAN_REPOSITORY,
         package_type: BakeRequest.PackageType.DEB.packageType,
-        packages: fullyQualifiedPackageName,
+        packages: unqualifiedPackageName,
```
duftler: What is the reason that this particular `package_name` parameter now results in an unqualified package name being returned?
It looks like it's just because of this change: `1 * imageNameFactoryMock.buildPackagesParameter(osPackages) >> unqualifiedPackageName`
But why is that change made to the particular test case?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandlerSpec.groovy
```diff
@@ -503,8 +545,9 @@ class GCEBakeHandlerSpec extends Specification {
       gceBakeHandler.producePackerCommand(REGION, bakeRequest)
 
     then:
-    1 * imageNameFactoryMock.deriveImageNameAndAppVersion(bakeRequest, _) >> [targetImageName, null, PACKAGES_NAME]
-      IllegalArgumentException e = thrown()
+    1 * imageNameFactoryMock.buildImageName(bakeRequest, osPackages) >> targetImageName
```
duftler: These 3 lines should be indented 2 more spaces.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -124,13 +133,17 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "reno-server",
                                         base_os: "ubuntu")
 
+    def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      print(imageName)
```
duftler: Extraneous?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -279,13 +318,17 @@ class DefaultImageNameFactorySpec extends Specification {
...
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced-0.1-3.all kato redis-server",
                                         base_os: "centos")
+      def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
+        PackageNameConverter.parseRpmPackageName(packageName)
+      }
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.head())
```
duftler: Missing a space before the `=`.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -259,15 +294,19 @@ class DefaultImageNameFactorySpec extends Specification {
...
                                         commit_hash: "170cdbd",
                                         base_os: "centos")
+      def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
+        PackageNameConverter.parseRpmPackageName(packageName)
+      }
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.head())
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -240,14 +272,17 @@ class DefaultImageNameFactorySpec extends Specification {
...
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced-0.1-3.all",
                                         build_number: "12",
                                         base_os: "centos")
 
+      def osPackage = PackageNameConverter.parseRpmPackageName(bakeRequest.package_name)
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -220,15 +249,18 @@ class DefaultImageNameFactorySpec extends Specification {
...
                                         build_number: "12",
                                         commit_hash: "170cdbd",
                                         base_os: "centos")
 
+      def osPackage = PackageNameConverter.parseRpmPackageName(bakeRequest.package_name)
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -202,13 +228,16 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         base_os: "ubuntu")
 
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -182,15 +204,19 @@ class DefaultImageNameFactorySpec extends Specification {
...
                                         commit_hash: "170cdbd",
                                         base_os: "ubuntu")
+    def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
+      PackageNameConverter.parseDebPackageName(packageName)
+    }
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.head())
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -162,15 +180,19 @@ class DefaultImageNameFactorySpec extends Specification {
...
                                         commit_hash: "170cdbd",
                                         base_os: "ubuntu")
+      def osPackages = bakeRequest.package_name.split(" ").collect { packageName ->
+        PackageNameConverter.parseDebPackageName(packageName)
+      }
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackages.head())
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -104,15 +111,17 @@ class DefaultImageNameFactorySpec extends Specification {
...
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         base_os: "ubuntu",
                                         ami_name: "nflx-base",
                                         ami_suffix: "1.0")
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+    def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+    def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -85,14 +90,16 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         base_os: "ubuntu",
                                         ami_suffix: "1.0")
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
 
     when:
-     def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -66,14 +69,16 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         base_os: "ubuntu",
                                         ami_name: "nflx-base")
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -47,14 +48,16 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         build_number: "12",
                                         base_os: "ubuntu")
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -16,26 +16,27 @@
...
       def bakeRequest = new BakeRequest(package_name: "nflx-djangobase-enhanced_0.1-3_all",
                                         build_number: "12",
                                         commit_hash: "170cdbd",
                                         base_os: "ubuntu")
+      def osPackage = PackageNameConverter.parseDebPackageName(bakeRequest.package_name)
 
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, [osPackage])
+      def appVersionStr= imageNameFactory.buildAppVersionStr(bakeRequest, osPackage)
```
duftler: Missing a space before the =.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverterSpec.groovy
```diff
@@ -99,7 +106,7 @@ class PackageNameConverterSpec extends Specification {
         parsedRpmPackageName)
 
     then:
-      parsedDebPackageName == parsedRpmPackageName
+      parsedDebPackageName != parsedRpmPackageName
```
duftler: Seems like we could compare each of the parts here to show that everything matches except for the `qualifiedName`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -140,7 +141,21 @@ abstract class CloudProviderBakeHandler {
...
+      PackageNameConverter.buildOsPackageName(selectedOptions.baseImage.packageType, name)
+    }
+
+    if (!osPackages.empty) {
```
duftler: Can typically just rely on Groovy truthiness when operating on a list (http://mrhaki.blogspot.com/2009/08/groovy-goodness-tell-groovy-truth.html). So this could read: `if (osPackages) {`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
brujoand: Good point. I'll add a test for empty/null packageNameList and remove the guard on bakeRequest.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
brujoand: No, it makes sense to use ```osPackageNames``` here also

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
brujoand: Initially it made sense since I'm basically just refactoring out the existing behaviour which only uses the first package. But looking at it again I think it's a good idea to send inn the entire list as it might be useful for someone else, and it puts the logic of making this choice where it belongs; within the method which can be overridden.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
brujoand: Yeah, I agree. I aimed at reducing ``` either :? or ``` for readability but it ended up more cryptic I think. Simplifying this instead.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
brujoand: 👍

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
brujoand: Oh good one. Checking just ```it``` looks much better.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
brujoand: ah, using the getter for both now. I'm a bit surprised that the former worked since it's marked private. I guess it's a bug of sorts. https://issues.apache.org/jira/browse/GROOVY-3010

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
brujoand: We would need the versionDelimiter in the method. Having it as a part of the OsPackageName seems random. But passing it as a parameter makes more sense in that it keeps the OsPackageName agnostic to what PackageType it's derived from. I'll try that.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
brujoand: Oh wow. Thanks for pointing this out. I just read up on the spock framework and the right shift operator. I've misunderstood this: ``` 1 * imageNameFactoryMock.buildPackagesParameter(osPackages) >> unqualifiedPackageName ``` to mean ``` verify that this method should be called once and that it returns unqualifiedPackageName ``` when the return actually tells the mock what it should return. Sorry about that, I'll take a fresh look at the tests with this in mind.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverterSpec.groovy
brujoand: I pulled qualifiedName into a method, so going back to quals which makes more sense anyway.

brujoand:
I've addressed the comments and added tests for empty packages, a test util trait to up the DRY score and aligned the indentation for the test classes so that they are consistent.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
```diff
@@ -16,27 +16,63 @@
...
-   *   1) A derived image name (to be used for naming the image being baked).
-   *   2) The appversion string (to be used for tagging the newly-baked image).
-   *   3) The updated list of packages to be used for overriding the passed package list (this will be removed once the
-   *      temporary workaround described above is removed).
-   *
-   * This function is not required to return the same image name on multiple invocations with the same bake request.
-   *
-   * Returns [imageName, appVersionStr, packagesParameter].
+   * Default implementation of ImageNameFactory considers only package_name, a timestamp, and base_os.
+   * For more fine grained conventions, extend this class, override the required methods and alter your
```
duftler: Does this work as described here? Does each `CloudProviderBakeHandler` resolve the correct `@Bean` factory method from that provider's configuration class when autowiring the `ImageNameFactory`?
In any case, I think it would be easier to follow if `CloudProviderBakeHandler` defined an abstract `getImageNameFactory` method, instead of relying on the autowiring for this.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/TestDefaults.groovy
```diff
@@ -0,0 +1,20 @@
+package com.netflix.spinnaker.rosco.providers.util
+
+import com.netflix.spinnaker.rosco.api.BakeRequest
+
+trait TestDefaults {
```
duftler: Nice.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
```diff
@@ -124,13 +133,17 @@ class DefaultImageNameFactorySpec extends Specification {
...
+      def imageNameFactory = new ImageNameFactory(clock: clockMock)
       def bakeRequest = new BakeRequest(package_name: "reno-server",
                                         base_os: "ubuntu")
 
+    def osPackages = parseDebOsPackageNames(bakeRequest.package_name)
+
     when:
-      def (imageName, appVersionStr, packagesParameter) = imageNameFactory.deriveImageNameAndAppVersion(bakeRequest, selectedOptions)
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      print(imageName)
```
duftler: Extraneous `print()`.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/openstack/OpenstackBakeHandlerSpec.groovy
```diff
@@ -116,564 +115,582 @@ class OpenstackBakeHandlerSpec extends Specification {
...
-==> openstack: Waiting for image dan-packer-test-3 (image id: 1f28b46b-b36f-4b7c-bc34-40e2371886fa) to become ready...
-==> openstack: Terminating the source server: 732520f1-af79-45be-bcda-ea1089e80f0d ...
-==> openstack: Deleting temporary keypair: packer 573d433f-278f-3e22-d2ce-e5720f906cf8 ...
-Build 'openstack' finished.
-
-==> Builds finished. The artifacts of successful builds are:
-"""
-
-    Bake bake = openstackBakeHandler.scrapeCompletedBakeResults(REGION, "123", logsContent)
+      def logsContent = """
```
duftler: These multiline strings probably shouldn't be reformatted.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/openstack/OpenstackBakeHandlerSpec.groovy
```diff
@@ -116,564 +115,582 @@ class OpenstackBakeHandlerSpec extends Specification {
...
-==> openstack: Terminating the source server: 732520f1-af79-45be-bcda-ea1089e80f0d ...
-==> openstack: Deleting temporary keypair: packer 573d433f-278f-3e22-d2ce-e5720f906cf8 ...
-Build 'openstack' finished.
-
-==> Builds finished. The artifacts of successful builds are:
---> openstack: An image was created: 1f28b46b-b36f-4b7c-bc34-40e2371886fa
-"""
-
-    Bake bake = openstackBakeHandler.scrapeCompletedBakeResults(REGION, "123", logsContent)
+      def logsContent = """
```
duftler: These multiline strings shouldn't be reformatted.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandlerSpec.groovy
```diff
@@ -420,13 +440,15 @@ class GCEBakeHandlerSpec extends Specification {
       def imageNameFactoryMock = Mock(ImageNameFactory)
       def packerCommandFactoryMock = Mock(PackerCommandFactory)
       def fullyQualifiedPackageName = "nflx-djangobase-enhanced_0.1-h12.170cdbd_all"
+      def unqualifiedPackageName = "nflx-djangobase-enhanced"
```
duftler: Is this intentionally still `unqualified`? Here and in several other tests?

duftler:
This is looking really nice. LGTM once the final comments are addressed.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactory.groovy
brujoand: Good point. I realise that this is much more in line with the current use of the bakeHandler vs roscoConfig also, so I'm changing it.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/ImageNameFactorySpec.groovy
brujoand: Ah forgot about that one. 👍

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandlerSpec.groovy
brujoand: Nope, I ment to revert those; doing it now.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/openstack/OpenstackBakeHandlerSpec.groovy
brujoand: oh, yes those indents shouldn't have been made part of the actual string. I'll strip the margins so we can still keep the indentation.

brujoand:
I've got back in sync with master and addressed the rest of the comments. I kept new indentation on the multiline strings but strip it so we can have both readability and reliability.  Let me know if I should just leave it as it is in master. 

Also I'm pretty sure there are more correct ways of handling the ``` getImageNameFactory ``` methods, but with a private field and a getter in the BakeHandlers I could keep the ease of mocking the ImageNameFactory and still avoid relying on autowireing.

duftler:
LGTM.
I agree that the handling of `getImageNameFactory` could be better, but I'll see if I can come up with something after merging this.

PR #123, opened 2016-07-24
provider/openstack: support only identity v3 api

drmaas:
Also add support for build_info_url in image metadata.


duftler:
LGTM.

PR #126, opened 2016-08-04
provider/aws: Move share_with and copy_to behavior into new set of te…

duftler:
…mplates.
One can specify the new `aws-multi-ebs.json` or `aws-multi-chroot.json` templates if the `share_with` and `copy_to` behavior is desired.
If the AWS packer builder(s) had logic to ignore whitespace-only `ami_users` and `ami_regions` values, having these two almost-identical sets of templates wouldn't be necessary.

### DIFF
--- a/rosco-web/config/packer/aws-chroot.json
+++ b/rosco-web/config/packer/aws-chroot.json
@@ -17,22 +17,7 @@
     "package_type": "",
     "packages": "",
     "upgrade": "",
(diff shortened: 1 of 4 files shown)


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/1001.

kalymero:
LGTM! Thanks for taking a look to it

PR #127, opened 2016-08-05
Adding custom naming conventions for docker provider

brujoand:
The thought here was to simply enable more relevant naming conventions for docker images. This took a bit more time than expected because I ended up fighting with the use of the extended_attributes. Now I'm at a point where I need some feedback to nail this. 

The problem, as I see it, is that the way that extended_attributes are currently handled we end up with a situation where the client of the Rosco API has to know how Rosco is implemented to not get in the way of it's job. 
In addition the CloudProvider-specific BakeHandler has to use a different extended_attribute name than the one being sent to packer in the ```parameterMap```. For instance I wanted to use ``` docker_target_image ``` as an extended attribute, which I would adapt to the naming convention in ``` DockeImageNameFactory ``` and then pass along to packer via the ``` parameterMap ```. However that was not possible due to the last instruction in ```CloudProviderBakeHandler.producePackerCommand``` which adds all extended_attributes to the parameterMap, and in effect overwriting anything which was previously there, except for ```share_with``` and ```copy_to``` which are handled manually. From my perspective the main problems with the current solution is that we end up with  'extra' attributes to avoid name collisions, and that you can't trust that entries in the ``` parameterMap ``` will not be overwritten. 

I ended up with the compromise where the CloudProvider-specific BakeHandles can choose which attributes are 'overridable' and promote those attributes in their ``` buildParameterMap``` method, using the ``` promoteExtendedAttributes ``` helper. This way overrides are  explicitly made possible, and extended_attributes which are not already present in the ``` parametrMap ``` are all added in ``` CloudProviderBakeHandler ``` like before.

Sorry for the wall of text. Wdyt?

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
@@ -120,7 +120,7 @@ abstract class CloudProviderBakeHandler {
   /**
    * Returns a map containing the parameters that should be propagated to the packer template for this provider.
    */
-  abstract Map buildParameterMap(String region, def virtualizationSettings, String imageName, BakeRequest bakeRequest)
+  abstract Map buildParameterMap(String region, def virtualizationSettings, String imageName, BakeRequest bakeRequest, String imageTag)
 
   /**
    * Returns the command that should be prepended to the shell command passed to the container.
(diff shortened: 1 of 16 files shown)


brujoand:
Initially I also let the docker_target_image_tag be overridden by the extended_attributes, but I think It's better to leave it with the default implementation, and rather use something like ``` docker_target_image_tags ``` in EA which could be a list and handled in the same manner as ```copy_to``` and ```share_with```. But I haven`t had the time to figure out how to tag with multiple tags in packer yet. 
Anyway, I got back in sync with master now.

brujoand:
Closing this and rewriting with more subtile approach.

PR #128, opened 2016-08-10
Default to the provided package name if package version and release c…

brujoand:
…an not be parsed

After #121 we found that there was a few cases where package names did not follow the appropriate naming conventions which resulted in a package name that failed to install.

This PR changes the parsing of package name in Rosco such that if we can not parse version and release, the package name will remain the same as the one provided in the BakeRequest.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
@@ -33,41 +33,42 @@ class PackageNameConverter {
     String arch
 
     public String qualifiedPackageName(PackageType packageType) {
-      if (!version || !release) return null // Not possible to generate without these
-
-      String versionDelimiter = packageType.getVersionDelimiter()
-      "${name}${versionDelimiter}${version}-${release}"
+      if (version && release) {
+        "${name}${packageType.getVersionDelimiter()}${version}-${release}"
+      } else if (version) {
+        "${name}${packageType.getVersionDelimiter()}${version}"
+      } else {
+        return null
+      }
     }
-
   }
 
   // Naming-convention for debs is name_version-release_arch.
   // For example: nflx-djangobase-enhanced_0.1-h12.170cdbd_all
   public static OsPackageName parseDebPackageName(String fullyQualifiedPackageName) {
     OsPackageName osPackageName = new OsPackageName()
+    if (!fullyQualifiedPackageName) return osPackageName
 
     osPackageName.with {
-      List<String> parts = fullyQualifiedPackageName?.tokenize("_")
+        name = fullyQualifiedPackageName
+        List<String> parts = fullyQualifiedPackageName?.tokenize("_")
+        if (parts) {
+          if (parts.size() > 1) {
+            List<String> versionReleaseParts = parts[1].tokenize("-")
+            if (versionReleaseParts) {
+              version = versionReleaseParts[0]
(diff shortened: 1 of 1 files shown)


rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverterSpec.groovy
```diff
@@ -34,8 +34,7 @@ class PackageNameConverterSpec extends Specification implements TestDefaults {
       null                                           | new PackageNameConverter.OsPackageName()
       ""                                             | new PackageNameConverter.OsPackageName()
       "nflx-djangobase-enhanced"                     | new PackageNameConverter.OsPackageName(name: "nflx-djangobase-enhanced")
-      "nflx-djangobase-enhanced_0.1"                 | new PackageNameConverter.OsPackageName(name: "nflx-djangobase-enhanced",
-                                                                                              version: "0.1")
+      "nflx-djangobase-enhanced_0.1"                 | new PackageNameConverter.OsPackageName(name: "nflx-djangobase-enhanced_0.1")
```
brujoand: As a result of this PR, having only version name is no longer enough to be parsed. But from what I can gather the BakeRequest value of the packageName would have been used in the previous versions. So in effect this should have the same result.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -45,29 +45,25 @@ class PackageNameConverter {
...
-        }
+      try {
+        List<String> parts = fullyQualifiedPackageName?.tokenize("_")
+        name = parts.get(0)
+
+        List<String> versionReleaseParts = parts.get(1).tokenize("-")
+        version = versionReleaseParts.get(0)
+        release = versionReleaseParts.get(1)
+        arch = parts[2]
+      } catch (RuntimeException e) {
```
duftler: Would rather we check the explicit conditions that might be false and then declare the naming-convention violated, instead of just catching any/all exceptions that might be thrown. Can still collect the extract the contents of the`catch` block into a utility method to be used at each spot we decide the name is invalid.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -77,27 +73,31 @@ class PackageNameConverter {
...
-        } else {
-          name = fullyQualifiedPackageName
-        }
+      try {
+        List<String> parts = packageName.tokenize("-")
+        if(parts.size() < 3) throw new RuntimeException("Parsing failed")
+        release = parts.pop()
+        version = parts.pop()
+        name = parts.join("-")
+      } catch (RuntimeException e) {
```
duftler: Same comment here.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
brujoand: Yeah, I failed to notice the ramifications of less data for the AppVersionStr. I'll try to rethink this.

brujoand:
I reverted most of the parsing and opted for keeping the ``` name ``` field as the original value up until we reach a point where we've parsed enough to be able to calculate the ``` qualifiedPackageName ```.  Wdyt?

PR #131, opened 2016-08-18
Add support for -var-file.

duftler:


duftler:
@lwander or @jtk54 please review, but do not merge yet.
I want to get the other pending rosco PR in first.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
```diff
@@ -64,6 +64,8 @@ class BakeRequest {
   String template_file_name
   @ApiModelProperty("A map of key/value pairs to add to the packer command")
   Map extended_attributes
+  @ApiModelProperty("The name of a json file containing key/value pairs to add to the packer command")
+  String var_file_name
```
lwander: Will these be listed in the UI?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
```diff
@@ -64,6 +64,8 @@ class BakeRequest {
   String template_file_name
   @ApiModelProperty("A map of key/value pairs to add to the packer command")
   Map extended_attributes
+  @ApiModelProperty("The name of a json file containing key/value pairs to add to the packer command")
```
lwander: Maybe make it clear these have to be in the config dir

lwander: Oh, actually looks like it's an absolute path. I guess that should be documented instead?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
duftler: The name of the file will be listed, but not the key/value pairs themselves. They are never populated into the context since they are only deserialized by packer itself.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
duftler: It's not an absolute path. It's treated the same as the template file name just above. I've amended the comment to make that clear.

jtk54:
LGTM

lwander:
LGTM

PR #132, opened 2016-08-23
Adding docker specific naming conventions

brujoand:
The thought here is to let ``` organization ``` become a full member of the bakeRequest, and let it, ``` ami_name ``` and ``` ami_suffix ``` provide the default basis for the ``` docker_target_image ```. 

We discussed (@duftler) letting the docker specific parameters like the docker image name come into the BakeRequest, but in retrospect it seems to me that re-using the ami_name/ami_suffix names, (and rather refactoring those some time in the future) would make more sense and keep the ``` BakeRequest ``` as uncluttered as possible. And I'm suggesting that organization might be general enough to be usefull for other providers, but this might be wrong? I think both ways work, so feel free to mix it up.  This PR will ofc also require a PR in Orca before the ``` organization ```  field can be utilised, which means we could also enrich that field which might be in line with what @lwander suggested earlier with regards to naming conventions and docker/kubernetes. FWIW I tried making this as simple as possible so these things can come step by step.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
@@ -1,0 +1,24 @@
+package com.netflix.spinnaker.rosco.providers.docker
+
+import com.netflix.spinnaker.rosco.api.BakeRequest
+import com.netflix.spinnaker.rosco.providers.util.ImageNameFactory
+import com.netflix.spinnaker.rosco.providers.util.PackageNameConverter
+
+class DockerImageNameFactory extends ImageNameFactory {
+
+  @Override
+  def buildAppVersionStr(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
+    super.buildAppVersionStr(bakeRequest, osPackageNames) ?: clock.millis().toString()
+  }
+
+  @Override
+  def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
+
+    String imageName = bakeRequest.ami_name ?: osPackageNames.first()?.name
+    String imageNamePrefixed = [bakeRequest.organization, imageName].findAll({it}).join("/")
+    String imageNameSuffixed = [imageNamePrefixed, bakeRequest.ami_suffix].findAll({it}).join("-")
+
+    return imageNameSuffixed
+  }
+
+}
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
@@ -155,6 +155,7 @@ class DockerBakeHandlerSpec extends Specification implements TestDefaults {
       def parameterMap = [
         docker_source_image: SOURCE_UBUNTU_IMAGE_NAME,
         docker_target_image: targetImageName,
+        docker_target_image_tag: SOME_MILLISECONDS,
         docker_target_repository: TARGET_REPOSITORY,
         repository: DEBIAN_REPOSITORY,
         package_type: DEB_PACKAGE_TYPE.packageType,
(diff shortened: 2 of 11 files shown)


lwander:
Great! Also `organization` is more of a Spinnaker thing than docker/kubernetes specific. BTW @duftler I thought we were trying to move away from the `ami_` specific naming?

brujoand:
> I thought we were trying to move away from the ami_ specific naming?

I think this was pretty much the conclusion from our chat yesterday, but I couldn't settle on a naming strategy general/specific. I guess ```image_name``` and ```image_name_suffix``` would be more suitable and give an okay migration path from ``` ami_name ``` and ``` ami_suffix ``` ?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  @Override
+  def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
+
+    String imageName = bakeRequest.ami_name ?: osPackageNames.first()?.name
+    String imageNamePrefixed = [bakeRequest.organization, imageName].findAll({it}).join("/")
+    String imageNameSuffixed = [imageNamePrefixed, bakeRequest.ami_suffix].findAll({it}).join("-")
+
+    return imageNameSuffixed
```
lwander: How will the tag be generated here?

lwander:
@brujoand I just talked to @duftler. I had misunderstood why he made the changes to Orca with Tomas in the first place, so changing stuff away from `ami_` now would just be a lot of painful work. Leaving it as is like you did sounds best.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
brujoand: If you mean the ```docker_image_tag``` the buildAppVersionStr is used by default in the DockerBakeHandler --> https://github.com/brujoand/rosco/blob/88a2bbce08caeb3d6c095adb6b904a8bd45713ec/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandler.groovy#L74 

I'm thinking that this would keep valuable information tagged to the image, and a next step could be to add support for multiple tags pr image from a user provided list. But as is, one could ofc put ``` docker_target_image_tag ``` in the extended attributes and override it.

brujoand:
Sounds good to me 👍

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactorySpec.groovy
```diff
@@ -0,0 +1,69 @@
...
+      def imageName = imageNameFactory.buildImageName(bakeRequest, osPackages)
+      def appVersionStr = imageNameFactory.buildAppVersionStr(bakeRequest, osPackages)
+      def packagesParameter = imageNameFactory.buildPackagesParameter(RPM_PACKAGE_TYPE, osPackages)
+    then:
+      imageName == "ECorp/nflx-djangobase-enhanced"
+      appVersionStr == "nflx-djangobase-enhanced-0.1-h12.170cdbd"
+      packagesParameter == "nflx-djangobase-enhanced-0.1-3 kato redis-server"
+  }
+
+  void "Should provide a docker image name based on ami_name, ami-suffix and no organization"() {
```
duftler: s/ami-suffix/ami_suffix

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
```diff
@@ -248,17 +251,16 @@ class DockerBakeHandlerSpec extends Specification implements TestDefaults {
...
     then:
       1 * imageNameFactoryMock.buildImageName(bakeRequest, osPackages) >> targetImageName
-      1 * imageNameFactoryMock.buildAppVersionStr(bakeRequest, osPackages) >> null
+      1 * imageNameFactoryMock.buildAppVersionStr(bakeRequest, osPackages) >> SOME_MILLISECONDS
       1 * imageNameFactoryMock.buildPackagesParameter(RPM_PACKAGE_TYPE, osPackages) >> PACKAGES_NAME
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, null, "$configDir/$dockerBakeryDefaults.templateFile")
   }
 
-  void 'produces packer command with all required parameters including appversion and build_host for trusty'() {
+  void 'produces packer command with all required parameters including build_host for trusty'() {
```
duftler: Maybe include in the test case description that the produced packer command includes `docker_target_image_tag` as well?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
duftler: @lwander Do you still have a question here?

duftler:
Nice clean approach.
LGTM after addressing comments.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
lwander: Sorry I forgot to respond - that looks great.

lwander:
LGTM2

brujoand:
Thanks,  I've update the test descriptions and synced with master.

PR #133, opened 2016-08-23
Remove instance type from openstack bakery request.

varikin:
Well, technically, don't check the instance type as it is defined by
region in the config. This approach is more inline with how AWS bakery
requests work.


duftler:
LGTM.

PR #134, opened 2016-08-23
provider/openstack: changes to packer to pass in image id instead of image name

drmaas:


drmaas:
@spinnaker/reviewers

rosco-web/config/rosco.yml
```diff
@@ -255,11 +255,11 @@ openstack:
       virtualizationSettings:
       - region: r1
         instanceType: smem-2vcpu
-        sourceImageName: ubuntu-latest
+        sourceImageId: b12f2467-cfca-4a65-b29b-d11b8c46818d
```
duftler: Would it make sense to drop a link here so a reader knows where this value is retrieved from? And does this have to be kept in sync somehow?

duftler:
LGTM after addressing comment.
Will need a rebase.

drmaas:
Link, as in comment?

Not sure what you mean about keeping it in sync - name and id are 2 of the attributes available for an image, and everything we're doing in spinnaker is operating on uuids.

duftler:
Yup, I meant as a comment.
W.r.t. keeping the id's in sync, I had something like this in mind: https://github.com/spinnaker/rosco/blob/master/rosco-web/config/rosco.yml#L52-L53
Are these id's placeholders, durable/will never change, or something else?

drmaas:
Added comment.

duftler:
Thanks.

PR #135, opened 2016-08-26
Install packer in docker for rosco (cloned from rush)

jcusa:
Noticed that packer is running from rosco instead of rush now. I wasn't able to bake ami successfully while  running rosco docker image.  Thus cloning similar change from rush docker to rosco docker. 

(https://github.com/spinnaker/rush/commit/f94712ad5fb2c4ae729c4ab8cc3a620f0903054a#diff-3254677a7917c6c01f55212f86c57fbf)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,4 +10,16 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test
 
 RUN dpkg -i ./rosco-web/build/distributions/*.deb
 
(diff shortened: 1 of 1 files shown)


duftler:
Thanks for the contribution.
LGTM.
@lwander please review as well.

lwander:
LGTM2

jcusa:
thank you,  @duftler , @lwander !!

PR #139, opened 2016-10-06
Adding a uuid for each bake request which is reused when running the …

brujoand:
…packer job. I'm also using this uuid when setting the docker tag so that we can trace a docker image back to a build. A later PR will extend the attributes returned when querying a bake result. 

The current naming conventions which uses the AppVersionStr broke whenever a slash was included in the tag, and this is a solution for that.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
@@ -31,6 +31,9 @@ import io.swagger.annotations.ApiModelProperty
 @CompileStatic
 class BakeRequest {
 
+  // A generated uuid which will identify the request and be used as the jobId when running the bake
+  @ApiModelProperty(value = "A generated UUID which will be used to identify the effective packer bake", readOnly = true)
+  final String request_id = UUID.randomUUID().toString()
   String user
   @ApiModelProperty("The package(s) to install") @JsonProperty("package") @SerializedName("package")
   String package_name
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
@@ -27,6 +27,7 @@ import groovy.transform.Immutable
 class JobRequest {
   List<String> tokenizedCommand
   List<String> maskedPackerParameters = []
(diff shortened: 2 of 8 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
```diff
@@ -27,6 +27,7 @@ import groovy.transform.Immutable
 class JobRequest {
   List<String> tokenizedCommand
   List<String> maskedPackerParameters = []
+  String jobId = ""
```
duftler: Is there a reason to initialize this to an empty string? I'd think there isn't any path hat has `JobRequest` being constructed without `jobId` being passed as a named constructor arg.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
```diff
@@ -31,6 +31,7 @@ import io.swagger.annotations.ApiModelProperty
 @CompileStatic
 class BakeRequest {
 
+  String request_id = UUID.randomUUID().toString()
```
duftler: Please add a comment here indicating that this arg is not expected to be passed in from the outside. Or if you consider it a reasonable mechanism for passing an external correlation id, describe that usage a bit.

Feels like this is intended for internal usage. If that is so, let's document it in a comment.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/JobRequest.groovy
duftler: Please add a comment explaining the relationship between this `jobId` and `BakeRequest.requestId`.

duftler:
Sorry, I hid the Comment button too soon. My comment was supposed to read: "Can you describe how you tested this PR to ensure there weren't negative implications to the other non-docker providers from this change?"

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
brujoand: 👍

brujoand:
Sorry for the delay. The cold viruses seemed intent on staying at our house this year, so I've been MIA for a while. I had addressed your comments earlier so I've just got back in touch with master, and pushed.

PR #140, opened 2016-10-10
Add new deleteBakeByKeyPreserveDetails() BakeStore entry-point and em…

duftler:
…ploy it when rebaking. This allows older bakes that have been rebaked to retain their logs and other details.
Modify deleteBakeByKey() to return the bake id of a deleted bake and include this in the response message.
Modify RedisBackedBakeStore.deleteBakeByKey() to resolve the appropriate rosco instance id instead of assuming the bake is being rebaked on the same rosco instance.
Modify BakeryController.createBake() to also cancel the executor's bake job when performing a rebake.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
@@ -50,6 +50,8 @@ class RedisBackedBakeStore implements BakeStore {
 
     jedis.withCloseable {
       scriptNameToSHAMap.with {
+        // Expected key list: lock key, bake key
+        // Expected arg list: ttlMilliseconds
         acquireBakeLockSHA = jedis.scriptLoad("""\
           -- Set the lock key key if it's not already set.
           if redis.call('SETNX', KEYS[1], 'locked') == 1 then
@@ -66,6 +68,8 @@ class RedisBackedBakeStore implements BakeStore {
             return false
           end
         """)
+        // Expected key list: "allBakes", bake id, bake key, this instance incomplete bakes key, lock key
+        // Expected arg list: createdTimestampMilliseconds, region, bake request json, bake status json, bake logs json, command, rosco instance id
         storeNewBakeStatusSHA = jedis.scriptLoad("""\
           -- Delete the bake id key.
           redis.call('DEL', KEYS[2])
@@ -108,11 +112,13 @@ class RedisBackedBakeStore implements BakeStore {
           -- Delete the lock key key instead of just allowing it to wait out the TTL.
           redis.call('DEL', KEYS[5])
         """)
+        // Expected key list: bake id
+        // Expected arg list: bake details json
         updateBakeDetailsSHA = jedis.scriptLoad("""\
           local existing_bake_status = redis.call('HGET', KEYS[1], 'bakeStatus')
 
           -- Ensure we don't update/resurrect a canceled bake (can happen due to a race).
-          if existing_bake_status and cjson.decode(existing_bake_status)["state"] == "CANCELED" then
+          if existing_bake_status and cjson.decode(existing_bake_status)['state'] == '$BakeStatus.State.CANCELED' then
             return
           end
 
@@ -125,11 +131,13 @@ class RedisBackedBakeStore implements BakeStore {
           -- Update the bake details set on the bake key hash.
           redis.call('HSET', bake_key, 'bakeDetails', ARGV[1])
         """)
+        // Expected key list: bake id, this instance incomplete bakes key
+        // Expected arg list: bake status json, bake logs json, updatedTimestampMilliseconds
         def updateBakeStatusBaseScript = """\
           local existing_bake_status = redis.call('HGET', KEYS[1], 'bakeStatus')
 
           -- Ensure we don't update/resurrect a canceled bake (can happen due to a race).
-          if existing_bake_status and cjson.decode(existing_bake_status)["state"] == "CANCELED" then
+          if existing_bake_status and cjson.decode(existing_bake_status)['state'] == '$BakeStatus.State.CANCELED' then
             return
           end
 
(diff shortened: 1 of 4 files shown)


duftler:
@lwander please review.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -176,17 +176,61 @@ class RedisBackedBakeStore implements BakeStore {
...
 
           if bake_id then
-            -- Remove the bake id from the set of incomplete bakes.
-            redis.call('SREM', KEYS[3], bake_id)
+            -- Retrieve the rosco instance id associated with bake id.
+            local rosco_instance_id = redis.call('HGET', bake_id, 'roscoInstanceId')
+
+            if rosco_instance_id then
+              -- Remove bake id from that rosco instance's set of incomplete bakes.
+              redis.call('SREM', 'allBakes:incomplete:' .. rosco_instance_id, bake_id)
```
lwander: What's this `..` syntax?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -176,17 +176,61 @@ class RedisBackedBakeStore implements BakeStore {
...
+          -- Delete the bake key key.
+          redis.call('DEL', KEYS[1])
+
+          if bake_id then
+            -- Retrieve the rosco instance id associated with bake id.
+            local rosco_instance_id = redis.call('HGET', bake_id, 'roscoInstanceId')
+
+            if rosco_instance_id then
+              -- Remove bake id from that rosco instance's set of incomplete bakes.
+              redis.call('SREM', 'allBakes:incomplete:' .. rosco_instance_id, bake_id)
```
lwander: Can you shed some light into what the different states & transitions are for a bake? I'm curious because the above `deleteBakeByKeySHA` script never sets the bake to canceled, but this one does.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -176,17 +176,61 @@ class RedisBackedBakeStore implements BakeStore {
...
+
+            -- Set bake state to CANCELED if still running.
+            local bake_status = redis.call('HGET', bake_id, 'bakeStatus')
+
+            if bake_status then
+              local bake_status_json = cjson.decode(bake_status)
+
+              if bake_status_json['state'] == 'RUNNING' then
+                bake_status_json['state'] = 'CANCELED'
+                bake_status_json['result'] = 'FAILURE'
```
lwander: It might be neater to pack these into a groovy enum and string-substitute them in.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -176,17 +176,61 @@ class RedisBackedBakeStore implements BakeStore {
...
+            end
 
             -- Delete the bake id key.
             redis.call('DEL', bake_id)
           end
 
-          return ret
+          return bake_id
+        """)
+        deleteBakeByKeyPreserveDetailsSHA = jedis.scriptLoad("""\
```
lwander: Could you annotate the required keys & args for each of these scripts?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -355,10 +399,34 @@ class RedisBackedBakeStore implements BakeStore {
...
+      Set<String> incompleteBakesKeys = jedis.keys(allIncompleteBakesKeyPattern)
+
+      keyList += incompleteBakesKeys
+    }
+
+    return evalSHA("deleteBakeByKeySHA", keyList, [])
+  }
+
+  @Override
+  public String deleteBakeByKeyPreserveDetails(String bakeKey) {
```
lwander: When will bake details be deleted if this path is taken?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
lwander: Ah, string concatenation

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -176,17 +176,61 @@ class RedisBackedBakeStore implements BakeStore {
...
 
           -- Delete the bake key key.
-          local ret = redis.call('DEL', KEYS[1])
+          redis.call('DEL', KEYS[1])
 
           if bake_id then
-            -- Remove the bake id from the set of incomplete bakes.
-            redis.call('SREM', KEYS[3], bake_id)
+            -- Retrieve the rosco instance id associated with bake id.
+            local rosco_instance_id = redis.call('HGET', bake_id, 'roscoInstanceId')
```
lwander: By deleting `bake_id` from `rosco_instance_ids` incomplete bakes, does that remotely cancel the bake?

duftler: Yup. The polling is done over each instance's incomplete bakes, as retrieved from that rosco instance's dedicated set of bake ids.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
duftler: The reason the one above never sets the bake to canceled is because the bake id mapping is deleted from redis in that script (that script is now only used for an explicit deletion of the bake and all its logs/details). In this script, we update it to canceled so someone querying it later won't think it's still running.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
duftler: Bake details aren't typically deleted. Only way (now) is via an explicit deletion call.
One key use case we are trying to enable here is the following:
  - Pipeline containing a bake executes. It results in a Bake execution details link being surfaced in the ui.
  - Later on, the requested image is 'rebaked'.
  - The original Bake execution details link should still function.

duftler:
`createBake` used to invoke `deleteBakeByKey()` if the `rebake` option was specified.
Whether on this rosco instance or another, any earlier bake would have had its logs/details deleted.
This pr also addresses a potential bug where the rebake request was handled by a rosco instance that is not the one currently handling an in-flight bake. The bake id would not have been removed from the correct incomplete bakes set.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
duftler: Done.

duftler:
All feedback addressed. PTAL.

PR #142, opened 2016-10-13
Add overloaded logs entry-point to return JSON based on the Accept: h…

duftler:
…eader on the request.


duftler:
The motivation for adding this is to allow gate to expose an entry-point for retrieving bakery logs. The most straightforward way to accomplish this (I think) is to have rosco return the logs to gate encoded via json since that is how gate is already configured to expect its responses.

duftler:
@lwander please review.

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -231,6 +232,22 @@ class BakeryController {
     throw new IllegalArgumentException("Unable to retrieve logs for '$statusId'.")
   }
 
+  @RequestMapping(value = "/api/v1/{region}/logs/{statusId}", produces = ["application/json"], method = RequestMethod.GET)
```
duftler: swagger seems to de-dupe based on the url pattern.

duftler:
Relates to spinnaker/spinnaker#832.

PR #146, opened 2016-11-01
Re-use newly exposed configurable polling settings.

duftler:


duftler:
@lwander please review.

PR #149, opened 2016-11-04
(openstack) Default username and password to empty string so /resolve…

duftler:
…dEnv works.


duftler:
fyi @drmaas and @varikin

PR #153, opened 2016-11-21
Handle metrics the same way as the other services (spectator).

duftler:


duftler:
@ewiseblatt please review.

PR #154, opened 2016-11-24
RFC: Extend the bake endpoint with build info

brujoand:
The rationale behind this PR is twofold;

1. We recently changed the defafult docker tag of baked images to be the bake id (#139), this PR will now allow us to ask rosco about a bake and get the build info based only on the docker tag. The alternative would be to find a way to incorporate this as metadata in the docker image, but this would mean we would have to find ways of doing this for all new cloud providers down the road. This PR instead relies on Rosco as a source of truth for this information. (Redis as a cache vs persistent storage might be an interesting topic here. )

2. Gard presented his ideas on general Artifacts last week, and it seemed to be an acceptance of the idea although more discussion might be needed for the actual implementation. But in this case it seemed logical to adapt to this general idea of an artifact when I had to change the return value of the bake endpoint anyway. So thoughts on that is also appreciated. 

This PR is first and foremost for discussion as we might want sturdier tests. It would also be worth a discussion wether the Artifact version should be a part of the Artifact object in the /bake endpoint. The /bake endpoint now returns something like this:

```
{
  "id" : "613bc102-9dad-42c3-bef2-c3ff4e484bb4",
  "ami" : null,
  "image_name" : "containers.schibsted.io/spt-infrastructure/mybuild:613bc102-9dad-42c3-bef2-c3ff4e484bb4",
  "artifact" : {
    "name" : null,
    "reference" : "containers.schibsted.io/spt-infrastructure/mybuild:613bc102-9dad-42c3-bef2-c3ff4e484bb4",
    "type" : "docker"
  },
  "build_info" : {
    "url" : null,
    "host" : "travis.schibsted.io",
    "job_name" : "jira-cli",
    "job_number" : 42,
    "organisation" : "spt-infrastructure"
  }
}
```

Which will let us get to the build info for docker images baked in Rosco which is currently not possible AFAIK. And it would probably also be helpfull to other providers.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Artifact.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Artifact.groovy
@@ -1,0 +1,12 @@
+package com.netflix.spinnaker.rosco.api
+
+/**
+ * The details of the baked artifact
+ *
+ * @see Bake
+ */
+class Artifact {
+  String name
+  String reference
+  String type
+}
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Bake.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Bake.groovy
@@ -32,6 +32,10 @@ import io.swagger.annotations.ApiModelProperty
 class Bake {
   @ApiModelProperty(value="The id of the bake job.")
   String id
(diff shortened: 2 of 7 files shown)


brujoand:
Now that we've added bake_url to the docker image, I'm not really sure this PR is needed anymore. If we want to use the artifacts (AMI/DockerImage/etc) as sources of truth instead of Rosco, we should probably just close this PR.

PR #156, opened 2016-12-11
Add installDist to buildRPM task for rosco-web

dmourati:


duftler:
Please rebase instead of merging master.

PR #157, opened 2016-12-14
Removing ami_suffix from docker image name

brujoand:
I had misunderstood the use of ami_suffix. And it seems unnatural to add it to the docker image name. For docker images I'm thinking that multiple tags / labels would be a better approach if we need such a qualifier.


rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactorySpec.groovy
```diff
@@ -28,7 +28,7 @@ class DockerImageNameFactorySpec extends Specification implements TestDefaults {
       packagesParameter == "nflx-djangobase-enhanced-0.1-3 kato redis-server"
   }
 
-  void "Should provide a docker image name based on ami_name, ami_suffix and no organization"() {
+  void "Should provide a docker image name based on ami_name and no organization"() {
```
duftler: Can probably drop the `ami_suffix` attribute from the `bakeRequest`.

brujoand: 👍

PR #158, opened 2016-12-14
Determine cloud provider for completed bakes via bake key instead of …

duftler:
…parsing first line of packer logs.


duftler:
@lwander please review.

PR #159, opened 2016-12-14
Updating base Ubuntu image for Azure

nzthiago:
@scotmoor @rguthriemsft @duftler PTAL

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -133,12 +133,12 @@ azure:
     baseImages:
     - baseImage:
         id: ubuntu
-        shortDescription: v14.04
-        detailedDescription: Ubuntu Server 14.04.4-LTS
+        shortDescription: v14.05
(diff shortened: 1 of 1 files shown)


rguthriemsft:
LGTM

rosco-web/config/rosco.yml
```diff
@@ -133,12 +133,12 @@ azure:
     baseImages:
     - baseImage:
         id: ubuntu
-        shortDescription: v14.04
-        detailedDescription: Ubuntu Server 14.04.4-LTS
+        shortDescription: v14.05
```
duftler: Looks like the minor number is still `04`, so this line should remain the same.

rguthriemsft: Understood it looks wrong, but when we checked VM skus using our Cli this was correct.

duftler:
Needs a rebase.

PR #162, opened 2016-12-20
Avoid bean collision with Boot 1.4.2 by not injecting ObjectMapper.

duftler:


duftler:
@ttomsu please review.

PR #165, opened 2017-01-08
Adds the ability to override the default templateFile for a baseImage

robumberger:


rosco-web/config/rosco.yml
```diff
@@ -59,6 +59,9 @@ aws:
         shortDescription: v12.04
         detailedDescription: Ubuntu Precise Pangolin v12.04
         packageType: deb
+        # You can specify the templateFile used for this baseImage
```
duftler: Please add punctuation to these two comments.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandlerSpec.groovy
```diff
@@ -145,7 +145,7 @@ class AzureBakeHandlerSpec extends Specification implements TestDefaults{
       AzureBakeHandler azureBakeHandler = new AzureBakeHandler(azureBakeryDefaults: azureBakeryDefaults)
 
     when:
-      String templateFileName = azureBakeHandler.getTemplateFileName()
+      String templateFileName = azureBakeHandler.getTemplateFileName(azureBakeHandler.bakeOptions.baseImages[0])
```
duftler: Would be nice if the `baseImages` had one that specified a `templateFile` so this test can be made to prove that the new override is respected.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -102,8 +102,9 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
     return parameterMap
   }
 
+
```
duftler: An extraneous blank line was added here.

PR #167, opened 2017-01-17
(gce) Add comments describing source image family configuration.

duftler:


duftler:
@edwinavalos please review.

rosco-web/config/rosco.yml
```diff
@@ -213,6 +213,10 @@ google:
       useInternalIp: false
       templateFile: gce.json
       baseImages:
+      # Either sourceImage or sourceImageFamily should be set. If both are set, sourceImage will take precedence.
+      # If you specify isImageFamily, deck will annotate the popup tooltip to indicate that the selected option
+      # represents an image family.
+      #
       # This is an example of configuring an explicit source image, as opposed to a source image family:
```
edwinavalos: Adding an example configuration like the below for the imageFamily settings (apart from what is being defaulted here) would be nice as well.

duftler: Good idea. Done.

PR #169, opened 2017-01-23
Fix instrumentation

ewiseblatt:
Renamed the spectator meter names because one was ambiguous and the other non-standard

bakesCompleted [Timer] formerly "bakes"   (was ambiguous)

bakesRequested [Counter] formerly "bakes" (was ambiguous)
  - changed the tag structure as well

bakesActive    [Gauge]   formerly "bakes.local"  (removed '.' from name and clarified)

@duftler

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
@@ -131,8 +131,9 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
 
                             // This will have the most up-to-date timestamp.
                             bakeStatus = bakeStore.retrieveBakeStatusById(statusId)
-                            Id failedBakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "orphanTimedOut")
-                            registry.timer(failedBakesId).record(bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp, TimeUnit.MILLISECONDS)
+                            def tags = [success: "false", cause: "orphanTimedOut", region: bakeStore.retrieveRegionById(statusId)]
+                            long millis = bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp
+                            registry.timer(registry.createId("bakesCompleted", tags)).record(millis, TimeUnit.MILLISECONDS)
                           }
                         }
                       },
@@ -157,14 +158,14 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
 
   void updateBakeStatusAndLogs(String statusId) {
     BakeStatus bakeStatus = executor.updateJob(statusId)
-    Id bakesId
+    def tags
 
     if (bakeStatus) {
       if (bakeStatus.state == BakeStatus.State.COMPLETED) {
         completeBake(statusId, bakeStatus.logsContent)
-        bakesId = registry.createId('bakes').withTag("success", "true")
+        tags = [success: "true"]
       } else if (bakeStatus.state == BakeStatus.State.CANCELED) {
-        bakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "jobFailed")
+        tags = [success: "false", cause: "jobFailed"]
       }
 
       bakeStore.updateBakeStatus(bakeStatus)
@@ -174,15 +175,17 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
       bakeStore.storeBakeError(statusId, errorMessage)
       bakeStore.cancelBakeById(statusId)
 
-      bakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "failedToUpdateJob")
+      tags = [success: "false", cause: "failedToUpdateJob"]
     }
 
-    if (bakesId) {
+    if (tags) {
       // This will have the most up-to-date timestamp.
       bakeStatus = bakeStore.retrieveBakeStatusById(statusId)
 
       if (bakeStatus) {
-        registry.timer(bakesId).record(bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp, TimeUnit.MILLISECONDS)
+        tags.region = bakeStore.retrieveRegionById(statusId)
+        long millis = bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp
+        registry.timer(registry.createId("bakesCompleted", tags)).record(millis, TimeUnit.MILLISECONDS)
       }
     }
   }
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
@@ -184,7 +184,8 @@ class JobExecutorLocal implements JobExecutor {
 
   @PostConstruct
   void initializeMetrics() {
-    registry.gauge(registry.createId("bakes.local").withTag("active", "true"), jobIdToHandlerMap, new ToDoubleFunction<Map>() {
+    // We need to have at least one tag.
+    registry.gauge(registry.createId("bakesActive", [active: "true"]), jobIdToHandlerMap, new ToDoubleFunction<Map>() {
 
       @Override
       double applyAsDouble(Map value) {
--- a/rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
+++ b/rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
@@ -16,7 +16,6 @@
 
 package com.netflix.spinnaker.rosco.controllers
 
-import com.netflix.spectator.api.Id
 import com.netflix.spinnaker.rosco.api.Bake
 import com.netflix.spinnaker.rosco.api.BakeOptions
 import com.netflix.spinnaker.rosco.api.BakeRequest
@@ -164,8 +163,7 @@ class BakeryController {
       def bakeKey = cloudProviderBakeHandler.produceBakeKey(region, bakeRequest)
 
       if (rebake == "1") {
-        Id bakesId = registry.createId('bakes').withTag("rebake", "true")
-        registry.counter(bakesId).increment()
+        registry.counter(registry.createId("bakesRequested", [flavor: "rebake"])).increment()
 
         String bakeId = bakeStore.deleteBakeByKeyPreserveDetails(bakeKey)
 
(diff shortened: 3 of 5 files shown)


ewiseblatt:
hold on, this isnt working in the runtime.

ewiseblatt:
ready.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -131,8 +131,9 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
 
                             // This will have the most up-to-date timestamp.
                             bakeStatus = bakeStore.retrieveBakeStatusById(statusId)
-                            Id failedBakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "orphanTimedOut")
-                            registry.timer(failedBakesId).record(bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp, TimeUnit.MILLISECONDS)
+                            def tags = ["success":"false", "cause":"orphanTimedOut", "region":bakeStore.retrieveRegionById(statusId)]
```
duftler: Please put a space after each colon.

duftler: Also don't need quotes around the map keys:

`[success: "false", cause: "orphanTimedOut", ...]`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -157,14 +158,14 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
...
   void updateBakeStatusAndLogs(String statusId) {
     BakeStatus bakeStatus = executor.updateJob(statusId)
-    Id bakesId
+    def tags = [:]
 
     if (bakeStatus) {
       if (bakeStatus.state == BakeStatus.State.COMPLETED) {
         completeBake(statusId, bakeStatus.logsContent)
-        bakesId = registry.createId('bakes').withTag("success", "true")
+        tags = ["success":"true"]
```
duftler: `tags.success = "true"`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -157,14 +158,14 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
...
       } else if (bakeStatus.state == BakeStatus.State.CANCELED) {
-        bakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "jobFailed")
+        tags = ["success":"false", "cause":"jobFailed"]
```
duftler: Could  assign each value individually. Not sure why we need to initialize `tags` up above if we're just overwriting it in each branch.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -174,15 +175,17 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
       bakeStore.storeBakeError(statusId, errorMessage)
       bakeStore.cancelBakeById(statusId)
 
-      bakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "failedToUpdateJob")
+      tags = ["success":"false", "cause":"failedToUpdateJob"]
     }
 
-    if (bakesId) {
+    if (!tags.isEmpty()) {
```
duftler: Any reason to not just use `!tags`? It will return false if there are no mappings.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -174,15 +175,17 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
...
       // This will have the most up-to-date timestamp.
       bakeStatus = bakeStore.retrieveBakeStatusById(statusId)
 
       if (bakeStatus) {
-        registry.timer(bakesId).record(bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp, TimeUnit.MILLISECONDS)
+        tags["region"] = bakeStore.retrieveRegionById(statusId)
```
duftler: `tags.region`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
```diff
@@ -174,15 +175,17 @@ class BakePoller implements ApplicationListener<ContextRefreshedEvent> {
...
+        long millis = bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp
+        registry.timer(registry.createId("bakesCompleted", tags)).record(millis, TimeUnit.MILLISECONDS)
```
duftler: I like the way you handled the tags here. Much nicer.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
```diff
@@ -184,7 +184,8 @@ class JobExecutorLocal implements JobExecutor {
 
   @PostConstruct
   void initializeMetrics() {
-    registry.gauge(registry.createId("bakes.local").withTag("active", "true"), jobIdToHandlerMap, new ToDoubleFunction<Map>() {
+    // We need to have at least one tag.
+    registry.gauge(registry.createId("bakesActive", ["active":"true"]), jobIdToHandlerMap, new ToDoubleFunction<Map>() {
```
duftler: `[active: "true"]`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -164,8 +163,7 @@ class BakeryController {
       def bakeKey = cloudProviderBakeHandler.produceBakeKey(region, bakeRequest)
 
       if (rebake == "1") {
-        Id bakesId = registry.createId('bakes').withTag("rebake", "true")
-        registry.counter(bakesId).increment()
+        registry.counter(registry.createId("bakesRequested", ["flavor":"rebake"])).increment()
```
duftler: `[flavor: "rebake"]`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -176,10 +174,11 @@ class BakeryController {
         def existingBakeStatus = queryExistingBakes(bakeKey)
 
         if (existingBakeStatus) {
-          Id bakesId = registry.createId('bakes').withTag("duplicate", "true")
-          registry.counter(bakesId).increment()
+          registry.counter(registry.createId("bakesRequested", ["flavor":"duplicate"])).increment()
```
duftler: `[flavor: "duplicate"]`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -176,10 +174,11 @@ class BakeryController {
...
 
           return existingBakeStatus
+        } else {
+          registry.counter(registry.createId("bakesRequested", ["flavor":"plain"]))
```
duftler: `[flavor: "plain"]`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -305,8 +304,8 @@ class BakeryController {
 
       // This will have the most up-to-date timestamp.
       BakeStatus bakeStatus = bakeStore.retrieveBakeStatusById(statusId)
-      Id failedBakesId = registry.createId('bakes').withTag("success", "false").withTag("cause", "explicitlyCanceled")
-      registry.timer(failedBakesId).record(bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp, TimeUnit.MILLISECONDS)
+      long millis = bakeStatus.updatedTimestamp - bakeStatus.createdTimestamp
+      registry.timer(registry.createId("bakesCompleted", ["success":"false", "cause":"explicitlyCanceled"])).record(millis, TimeUnit.MILLISECONDS)
```
duftler: `[success: "false", cause: "explicitlyCanceled"]`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
duftler: 'plain' sounds a little weird. Maybe 'new' or something like that? I don't feel that strongly about it, but it isn't very descriptive.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
ewiseblatt: Are you sure? It is a simpler line, but isnt as clear. It suggests to me that there could be other [default] tags.

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
ewiseblatt: "new" sounds like "rebake" (as in I want a new bake no matter what). I'd call "rebake" "force" except that you call it rebake so think that terminology might be used in practice and didnt want to introduce new terms.

I could use default, though default is even less descriptive. I was thinking "plain" because the request isnt decorated (not rebake). "duplicates" happen to be a variant of "plain" though that isnt evident. And "plain" and "flavor" went together. 

What about "normal"?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
duftler: Ok, fair point.

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
duftler: I think I like 'plain' better than 'normal'. Let's just stick with 'plain' for now.

ewiseblatt:
When i fix the other issue about tags being declared empty up front, this
becomes moot.
I changed it so tags is null and I check for null below. So this makes more
sense anyway.

ewiseblatt:
Look at the tests. Make sure the counters are as you expected.
In particular where there are errors and failures.
There are no failure counters, however the controller.invocation interceptor would catch those failures. I dont know if you want a more direct and tied counter here.

PR #170, opened 2017-01-28
(google) Add support for subnetwork config default and gce_subnetwork…

duftler:
… extended attribute.


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/1391.

duftler:
@lwander or @danielpeach please review.

rosco-web/config/packer/gce.json
```diff
@@ -4,6 +4,7 @@
     "gce_account_file": "",
     "gce_zone": null,
     "gce_network": null,
+    "gce_subnetwork": "",
```
lwander: Why is subnetwork initialized with `""` rather than `null` like the other `String` parameters above?

duftler: `null` means required, `""` means optional.

PR #171, opened 2017-01-29
(AWS/Azure) Add support for Windows using Chocolatey as package manager.

robumberger:
There are 2 PR's that need to be merged before this PR can be merged. 

- Spinnaker: [PR #1390](https://github.com/spinnaker/spinnaker/pull/1390)
- Orca: [PR #1146](https://github.com/spinnaker/orca/pull/1146)

I've also raised an issue with Packer as when using the powershell provisioner via Spinnaker, the powershell command line isn't properly formed. Issuing the exact same packer command line via the command line works as expected. Please note that this only impacts Azure. AWS is working as expected.

- Packer issue: [Issue #4478](https://github.com/mitchellh/packer/issues/4478)

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
@@ -26,7 +26,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
     def packerCommand = [baseCommand, "packer", "build", "-color=false"]
     parameterMap.each { key, value ->
       if (key && value) {
-        def keyValuePair = value instanceof String && value.contains(" ") ? "$key=\"$value\"" : "$key=$value"
+        def keyValuePair = "$key=${value instanceof String ? value.trim() : value}"
 
         packerCommand << "-var"
         packerCommand << keyValuePair.toString()
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
@@ -106,12 +106,101 @@ class PackageNameConverter {
     osPackageName
   }
 
+  // Nuget supports SemVer 1.0 for package versioning standards.
+  // Therefore, the naming convention for nuget packages is
+  // {package-name}[.{language}].{major}.{minor}.{patch/build}[-{version}][.{revision}][+{metadata}]
+  // For example: ContosoUtilities.ja-JP.1.0.0-rc1.nupkg
+  //              notepadplusplus.7.3.nupkg
+  //              autohotkey.1.1.24.04.nupkg
+  //              microsoft-aspnet-mvc.6.0.0-rc1-final.nupkg
+  //              microsoft-aspnet-mvc.de.3.0.50813.1.nupkg
+  //              microsoft.aspnet.mvc.6.0.0-rc1-final.nupkg
+  //              microsoft.aspnet.mvc.de.3.0.50813.1.nupkg
+  public static OsPackageName parseNupkgPackageName(String fullyQualifiedPackageName) {
+    OsPackageName osPackageName = new OsPackageName()
+    if (!fullyQualifiedPackageName) return osPackageName
+
+    fullyQualifiedPackageName = fullyQualifiedPackageName.replaceFirst('.nupkg', '')
+
+    osPackageName.with {
+      name = fullyQualifiedPackageName
+
+      List<String> parts = fullyQualifiedPackageName.tokenize(".")
+
+      if (parts.size() > 2) {
+        def versionStart = 0
+
+        for (def i = 0; i < parts.size(); i++) {
+          if (i > 0 && parts[i].isInteger()) {
+            versionStart = i
+            break
+          }
+        }
+
+        if (versionStart < 1) {
+          for (def i = 0; i < parts.size(); i++) {
+            if (i > 0 && parts[i].contains('-') && parts[i][0].isInteger()) {
+              versionStart = i
+              break
+            }
+          }
+        }
+
+        if (versionStart > 0) {
+
+          name = parts.subList(0, versionStart).join('.')
+          version = parts.subList(versionStart, parts.size()).join('.')
+
+          if (version.contains('-')) {
+            (version, release) = version.split('-', 2)
+          } else if (version.contains("+")) {
+            (version, release) = version.split("\\+", 2)
+            release = '+' + release
+          }
+
+        } else {
+
+          def metaDataIndex = parts.findIndexOf { val -> val =~ /\+/ }
+
+          if (metaDataIndex > -1) {
+            (name, release) = parts[metaDataIndex].split('\\+')
+            release = "+" + release
+            name = parts.subList(0, metaDataIndex).join('.') + "." + name
+          } else {
+            name = parts.join('.')
+          }
+        }
+
+      } else if (parts.size() == 2) {
+
+        if (parts[1].isInteger()) {
+          name = parts[0]
+          version = parts[1]
+        } else if (parts[1][0].isInteger() && parts[1].contains('-')) {
+
+            name = parts[0]
+
+            def versionParts = parts[1].split('-', 2)
+            version = versionParts[0]
+            release = versionParts[1]
+
+        } else {
+          name = parts.join(".")
+        }
+      }
+    }
+
+    osPackageName
+  }
+
   public static OsPackageName buildOsPackageName(BakeRequest.PackageType packageType, String packageName) {
     switch (packageType) {
       case BakeRequest.PackageType.DEB:
         return PackageNameConverter.parseDebPackageName(packageName)
       case BakeRequest.PackageType.RPM:
         return PackageNameConverter.parseRpmPackageName(packageName)
+      case BakeRequest.PackageType.NUPKG:
+        return PackageNameConverter.parseNupkgPackageName(packageName)
       default:
         throw new IllegalArgumentException("Unrecognized packageType '$packageType'.")
     }
@@ -123,22 +212,26 @@ class PackageNameConverter {
     }.findAll{it.name}
   }
 
-  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName) {
+  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName, BakeRequest.PackageType packageType) {
     // As per source of AppVersion, these are valid appversion tags:
     //   subscriberha-1.0.0-h150
     //   subscriberha-1.0.0-h150.586499
     //   subscriberha-1.0.0-h150.586499/WE-WAPP-subscriberha/150
     String appVersion = osPackageName.name
 
+    def versionSeparator = packageType == BakeRequest.PackageType.NUPKG ? "." : "-"
+    def buildNumberSeparator = packageType == BakeRequest.PackageType.NUPKG ? "-" : "-h"
+    def commitHashSeparator = packageType == BakeRequest.PackageType.NUPKG ? "+" : "."
+
     osPackageName.with {
       if (version) {
-        appVersion += "-$version"
+        appVersion += "$versionSeparator$version"
 
         if (bakeRequest.build_number) {
-          appVersion += "-h$bakeRequest.build_number"
+          appVersion += "$buildNumberSeparator$bakeRequest.build_number"
 
           if (bakeRequest.commit_hash) {
-            appVersion += ".$bakeRequest.commit_hash"
+            appVersion += "$commitHashSeparator$bakeRequest.commit_hash"
           }
 
           if (bakeRequest.job) {
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
@@ -37,6 +37,7 @@ class AWSBakeHandlerSpec extends Specification implements TestDefaults {
   private static final String SOURCE_UBUNTU_PV_IMAGE_NAME = "ami-a654321b"
   private static final String SOURCE_TRUSTY_HVM_IMAGE_NAME = "ami-c456789d"
   private static final String SOURCE_AMZN_HVM_IMAGE_NAME = "ami-8fcee4e5"
+  private static final String SOURCE_WINDOWS_2012_R2_HVM_IMAGE_NAME = "ami-21414f36"
 
   @Shared
   String configDir = "/some/path"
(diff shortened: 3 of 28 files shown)


robumberger:
Packer issue 4478 has been solved

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
```diff
@@ -26,7 +26,18 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
     def packerCommand = [baseCommand, "packer", "build", "-color=false"]
     parameterMap.each { key, value ->
       if (key && value) {
-        def keyValuePair = value instanceof String && value.contains(" ") ? "$key=\"$value\"" : "$key=$value"
+        def cleansedValue = value instanceof String ? value.trim() : value
```
duftler: This change originally concerned me, but I think it's ok. I ran a test against gce using a space-separated list of package names and it was handled properly. I _think_ this may have been some historical logic left over from when we were doing builds using rush & containers (I also recall that we were tokenizing the command on spaces, I _think_ just before sending it to rush, but that was a while back and the memories are fading).

Also, why do we need to handle the case where the param has leading/trailing quotes? Is there some path that creates that situation (other than a user actually adding those quotes into the packages field of the bake stage ui; I'd think quoting values in just about any of the fields throughout the ui would create problems)? If there isn't a path that leads to having those surrounding quotes, let's just drop the decorating & cleansing logic altogether.

@gardleopard Any concerns with this change?

rosco-web/config/rosco.yml
```diff
@@ -5,7 +5,9 @@ rosco:
   configDir: /some/path/to/config/packer
   jobs:
     local:
-      timeoutMinutes: 20
+      timeoutMinutes: 30
```
robumberger: In Azure, my Windows bakes were fairly consistently timing out at the 20 minute mark. The Windows bakes usually took 20-25 minutes with 2 Chocolatey packages (Chrome and Java 8). I bumped up to 30 minutes to allow for enabling Windows features since these can take time to do. "orphanedJobTimeoutMinutes" was increased by 5 minutes to account for the extra bake time. Having "timeoutMinutes" and "orphanedJobTimeoutMinutes" set to 30 minutes would end up as a bit of a race condition.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -123,29 +212,53 @@ class PackageNameConverter {
     }.findAll{it.name}
   }
 
-  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName) {
+  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName, BakeRequest.PackageType packageType) {
     // As per source of AppVersion, these are valid appversion tags:
     //   subscriberha-1.0.0-h150
     //   subscriberha-1.0.0-h150.586499
     //   subscriberha-1.0.0-h150.586499/WE-WAPP-subscriberha/150
+    //   ContosoUtilities.ja-JP.1.0.0-rc1+asdfe
```
robumberger: I've removed that comment as it shouldn't have been included. I'll check the link and verify. We haven't seen any issues yet but doesn't mean they don't exist.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -123,29 +212,53 @@ class PackageNameConverter {
...
     String appVersion = osPackageName.name
 
-    osPackageName.with {
-      if (version) {
-        appVersion += "-$version"
+    if (packageType == BakeRequest.PackageType.NUPKG) {
```
robumberger: Fair statement. Will update accordingly.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -308,11 +365,51 @@ class AWSBakeHandlerSpec extends Specification implements TestDefaults {
 
     then:
       1 * imageNameFactoryMock.buildImageName(bakeRequest, osPackages) >> targetImageName
-      1 * imageNameFactoryMock.buildAppVersionStr(bakeRequest, osPackages) >> null
+      1 * imageNameFactoryMock.buildAppVersionStr(bakeRequest, osPackages, RPM_PACKAGE_TYPE) >> null
       1 * imageNameFactoryMock.buildPackagesParameter(RPM_PACKAGE_TYPE, osPackages) >> PACKAGES_NAME
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, null, "$configDir/$awsBakeryDefaults.templateFile")
   }
 
+  void 'produces packer command with all required parameters for amzn, using default vm type, with debug flag set'() {
```
robumberger: No difference and will remove. I initially had code included to output the debug details from Packer as part of the bake log, but moved that to a different branch for later submission. I must have missed this test when cleaning up.

rosco-web/config/rosco.yml
duftler: There wouldn't be any effect on the bake from leaving `orphanedJobTimeoutMinutes` as it was. That value is only compared against the amount of time since the in-flight bake was last updated. If a rosco instance is managing the bake it will be updated every 15 seconds. That is, it compares the 'staleness' of a bake against this threshold, not the lifespan of a bake.

https://github.com/spinnaker/rosco/blob/c69d724f9cdc581a9bd60c19d649272c0eb31087/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy#L117-L122

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
robumberger: This change was due to the Packer issue I reported ([#4478](https://github.com/mitchellh/packer/issues/4478)). The issue arose when there were 2 or more packages in the packages variable that was being sent to packer. When building the packer command, the value of the "packages" parameter included quotes, which were then being included in the escaped quotes due to the space. This quotes within quotes was being included in the Powershell script which caused a parsing issue. The root issue might be with the way Apache Commons Exec handles quoted quotes (there's a related [issue](https://issues.apache.org/jira/browse/EXEC-54) open since 2012 that seems to relate), but this approach was easier. One of the packer contributors worked on this issue with me and confirmed that the way this now passes it's command line arguments to packer is okay even if the quotes are missing.

rosco-web/config/rosco.yml
robumberger: Got it, will revert that change

rosco-web/config/rosco.yml
```diff
@@ -171,6 +187,18 @@ azure:
...
         packageType: rpm
+    - baseImage:
+        id: windows-2012-r2
+        shortDescription: 2012 R2
+        detailedDescription: Windows Server 2012 R2 Datacenter
+        publisher: MicrosoftWindowsServer
+        offer: WindowsServer
+        sku: 2012-R2-Datacenter
+        version: 4.0.20170111
+        osType: Windows
```
duftler: The `windows` up above is lower-case, but this one is capitalized. I don't have a preference so long as each style works. But we should just pick one.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.groovy
```diff
@@ -123,22 +212,26 @@ class PackageNameConverter {
...
 
-  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName) {
+  public static String buildAppVersionStr(BakeRequest bakeRequest, OsPackageName osPackageName, BakeRequest.PackageType packageType) {
     // As per source of AppVersion, these are valid appversion tags:
     //   subscriberha-1.0.0-h150
     //   subscriberha-1.0.0-h150.586499
     //   subscriberha-1.0.0-h150.586499/WE-WAPP-subscriberha/150
     String appVersion = osPackageName.name
 
+    def versionSeperator = packageType == BakeRequest.PackageType.NUPKG ? "." : "-"
```
duftler: s/Seperator/Separator for these 3 lines.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
brujoand: I can confirm the change on line 29 being ok. We actually did this change locally a while a go, but never got around to push it upstream. We've also stopped removing leading/trailing space in the `install_packages.sh` script, so I'm curious as to why it is needed here? And if it's needed I think that `trim()` would handle this case, right?

robumberger: Removal of the extra leading and trailing quotes are needed as when passing the "packages" parameter to packer. I'm very new to Groovy, but my understanding is that trim() doesn't do that.

The command line that was being generated by Spinnaker was `packer build ... -var packages="googlechrome javaruntime" azure-windows-2012-r2.json`. If you ran this same command line via an interactive shell and then looked at /proc/<proces_id>/cmdline, you'll see it as:

```
ubuntu@spinnaker:~$ cat /proc/7472/cmdline
/usr/bin/packerbuild-color=false-debug-varazure_client_id=##########ClientID##################-varazure_client_secret=#########-varazure_resource_group=spinnaker-varazure_storage_account=spinnakerimages-varazure_subscription_id=########SubscriptionID##############-varazure_tenant_id=##########TenantID##################-varazure_object_id=##########ObjectID##################-varazure_location=eastus-varazure_image_publisher=MicrosoftWindowsServer-varazure_image_offer=WindowsServer-varazure_image_sku=2012-R2-Datacenter-varazure_image_name=null-null-varpackage_type=nupkg-varpackages=googlechrome javaruntime-varconfigDir=/opt/rosco/config/packer/opt/rosco/config/packer/azure-windows-2012-r2.json
```

This looks fine as you don't see the quotes. When Spinnaker runs packer via a non-interactive shell and you look at /proc/<proces_id>/cmdline, you'll see it as:

```
ubuntu@spinnaker:~$ cat /proc/7619/cmdline
/usr/bin/packerbuild-color=false-debug-varazure_client_id=##########ClientID##################-varazure_client_secret=#########-varazure_resource_group=spinnaker-varazure_storage_account=spinnakerimages-varazure_subscription_id=########SubscriptionID##############-varazure_tenant_id=##########TenantID##################-varazure_object_id=##########ObjectID##################-varazure_location=eastus-varazure_image_publisher=MicrosoftWindowsServer-varazure_image_offer=WindowsServer-varazure_image_sku=2012-R2-Datacenter-varazure_image_name=null-null-varpackage_type=nupkg-varpackages="googlechrome javaruntime"-varconfigDir=/opt/rosco/config/packer/opt/rosco/config/packer/azure-windows-2012-r2.json
```

You'll notice the extra quotes for the 'packages' variable, which causes packer to pass the parameter as `packages=""googlechrome javaruntime""` to Powershell which causes a parsing error.

duftler: @robumberger My original question was about quotes being passed _in_ to this method as part of the parameter value? I can't see any path that creates this. If there isn't any path that creates this situation, we don't need to trim surrounding quotes here; we just need to be sure that we don't add any more.

duftler: Btw, I read the linked packer issue and I agree with the reasoning there.

robumberger: @duftler I wasn't able to track down where the extra quotes were coming from, but haven't looked through the entire Spinnaker codebase yet to figure out how all of it fits together. I originally tried not trimming the quotes, but that didn't work hence my solution. If this is a blocking issue from getting the PR merged, I'll dig in some more this weekend and see if I can track it down. If it's not a blocking issue, I'll add it to my list of Spinnaker related items and will get to it in the next couple of weeks.

duftler: I think just adding some instrumentation to this method to `println` the value of `value` for each attribute, and then running whatever pipeline you've been running, should be sufficient. I never saw surrounding quotes here.

robumberger: I went back through the various changes I made previously while debugging this issue and couldn't reproduce the behavior I saw. Code is no longer stripping out the leading/trailing quotes.

rosco-web/config/rosco.yml
rguthriemsft: LGTM, my one comment is currently in azure bake stage we only support use of a single template.  We have a backlog to support multiple templates which I was about to do in next week or two.  WIll include you on that PR when submitted as I assume you need that.  @scotmoor @ericjizbmsft

rguthriemsft: @robumberger currently bake in packer does not require use of objectid, can you comment on why this is needed?  Also is this object id of the service principal?  @ericjizbamsft @scotmoor

robumberger: @rguthriemsft  When baking Windows images, we are using WinRM to connect to and provision the instance. For the Azure Resource Manager Builder, packer automatically creates the WinRM certificate for Windows machines. The object_id, which is the object id of the service principal, is required in order to protect the certificate. Please see the [Azure ARM](https://www.packer.io/docs/builders/azure-arm.html) docs which specify that this is a required parameter when the os_type is 'Windows'.

rguthriemsft: Ah got it, when we did packer support a lot of the packer stuff wasn't there so this makes sense.  As long as users not leverage windows image support are not required to add in object id, no objections to PR.

duftler: Ok, so is this resolved?
@rguthriemsft This LG2U? And the companion spinnaker/spinnaker pr as well?

rguthriemsft: @robumberger can you ack that you tested not setting this param with Linux bake?  @duftler once rob acks, we are good with pull request 171, 1390 and 1146.

robumberger: @rguthriemsft Confirmed - Linux bake tested while having 'object_id' defined and undefined in spinnaker-local.yml and no issues found. Linux does not pass 'object_id' to Packer in either case.

rguthriemsft: @robumberger thank you for confirmation and thank you for the contribution.  @duftler we are good, please merge.

PR #172, opened 2017-02-03
Adding docker tag, organization and name handling

brujoand:
This PR tries to let us use and control organization, tag and the image name for docker in a more coherent way while still respecting the current contract. My main motivation is to be able to enrich the bake step in deck for a separate PR.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
@@ -15,6 +15,7 @@ class DockerImageNameFactory extends ImageNameFactory {
   def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
 
     String imageName = bakeRequest.ami_name ?: osPackageNames.first()?.name
+
     String imageNamePrefixed = [bakeRequest.organization, imageName].findAll({it}).join("/")
 
     return imageNamePrefixed
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
@@ -150,6 +150,7 @@ class DockerBakeHandlerSpec extends Specification implements TestDefaults {
                                         package_name: PACKAGES_NAME,
                                         base_os: "ubuntu",
                                         request_id: SOME_UUID,
+                                        commit_hash: SOME_COMMIT_HASH,
                                         build_info_url: SOME_BUILD_INFO_URL,
                                         cloud_provider_type: BakeRequest.CloudProviderType.docker)
       def targetImageName = "kato-x8664-timestamp-ubuntu"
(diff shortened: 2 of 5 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
```diff
@@ -13,8 +13,12 @@ class DockerImageNameFactory extends ImageNameFactory {
 
   @Override
   def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
+    String imageName = bakeRequest.extended_attributes?.get('docker_target_image_name')
```
duftler: I know the name isn't perfect, but doesn't specifying `ami_name` accomplish the same thing as you're doing with `docker_target_image_name`?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
```diff
@@ -150,6 +150,7 @@ class DockerBakeHandlerSpec extends Specification implements TestDefaults {
                                         package_name: PACKAGES_NAME,
                                         base_os: "ubuntu",
                                         request_id: SOME_UUID,
+                                        commit_hash: SOME_COMMIT_HASH,
```
duftler: Might be good to add a test case demonstrating that `docker_target_image_tag` is always set to _something_. If there's no `commit_hash`, and no explicit `docker_target_image_name`, just verify that the tag was assigned some non-null value? Don't see any need to mock the clock or anything like that.

brujoand: Ah, good point. I'll add that 👍

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
brujoand: Yeah, I had a vague memory of us discussing this a while back. I seemed to remember that we wanted to avoid using the ami_name for other providers than AWS but I might be wrong. 

Anyway, I have no strong feelings on this so I don't mind using the ami_name if that's something that we want. Keep ami_name?

duftler: Since we're already using it here, I'd say just stick with it for now.

brujoand: Cool 👍

brujoand:
Relying on `ami_name` and added test for `docker_target_image_tag` defaulting to a long timestamp.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
```diff
@@ -15,6 +15,7 @@ class DockerImageNameFactory extends ImageNameFactory {
   def buildImageName(BakeRequest bakeRequest, List<PackageNameConverter.OsPackageName> osPackageNames) {
 
     String imageName = bakeRequest.ami_name ?: osPackageNames.first()?.name
+
```
duftler: Extraneous blank line?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerBakeHandlerSpec.groovy
```diff
@@ -363,6 +367,26 @@ class DockerBakeHandlerSpec extends Specification implements TestDefaults {
...
+      dockerVirtualizationSettings.sourceImage = SOURCE_UBUNTU_IMAGE_NAME
+
+    when:
+      def parameterMap = dockerBakeHandler.buildParameterMap(
+        REGION, dockerVirtualizationSettings, "myDockerImage", bakeRequest, "SomeVersion"
+      )
+
+    then:
+      // This should default to a timestamp of type long if we don't provide a tag or commit hash.
+      parameterMap['docker_target_image_tag'].toLong()
```
duftler: Nice.

PR #177, opened 2017-02-08
(Azure) Ensure 'azure_image_name' conforms to CaptureNamePrefix regex in packer

robumberger:
While testing bake-only pipelines, I noticed that the image name in Azure always had a prefix of "null-null". This is due to 'build_number' and 'base_name' not being defined. This PR will use the 'imageName' passed into the 'buildParameterMap' method if the 'build_number' and 'base_name' are not defined. It also ensures that the generated 'azure_image_name', which is passed to packer as 'capture_name_prefix', meets the packer naming standard for this field. 

The packer code defines a [regex](https://github.com/mitchellh/packer/blob/master/builder/azure/arm/config.go#L45) that limits the character set to alphanumeric characters, underscore, dash, and period. The value can also not exceed 23 characters, nor end with a period or dash.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
@@ -25,6 +25,8 @@ import com.netflix.spinnaker.rosco.providers.util.ImageNameFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
+import java.time.Clock
+
 @Component
 public class AzureBakeHandler extends CloudProviderBakeHandler{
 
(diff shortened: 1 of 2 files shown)


rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandlerSpec.groovy
```diff
@@ -331,4 +331,34 @@ class AzureBakeHandlerSpec extends Specification implements TestDefaults{
...
+      build_number: buildNumber,
+      base_name: baseName)
+    def azureVirtualizationSettings = azureBakeHandler.findVirtualizationSettings(REGION, bakeRequest)
+    def parameterMap = azureBakeHandler.buildParameterMap(REGION, azureVirtualizationSettings, imageName, bakeRequest, "SomeVersion")
+
+    then:
+    parameterMap.azure_image_name == expectedAzureImageName
+
+    where:
+    buildNumber   | baseName    | imageName                                           | expectedAzureImageName
```
duftler: Can use `||` instead of `|` just before `expectedAzureImageName`.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -92,10 +94,26 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
       azure_location: region,
       azure_image_publisher: selectedImage?.baseImage?.publisher,
       azure_image_offer: selectedImage?.baseImage?.offer,
-      azure_image_sku: selectedImage?.baseImage?.sku,
-      azure_image_name: "$bakeRequest.build_number-$bakeRequest.base_name"
+      azure_image_sku: selectedImage?.baseImage?.sku
     ]
 
+    if (bakeRequest?.build_number && bakeRequest?.base_name) {
```
duftler: Don't need the null-safe navigation, since `bakeRequest` can't be null here.

robumberger:
Thanks for the feedback. Requested changes have been made.

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandlerSpec.groovy
```diff
@@ -331,4 +331,34 @@ class AzureBakeHandlerSpec extends Specification implements TestDefaults{
...
+      build_number: buildNumber,
+      base_name: baseName)
+    def azureVirtualizationSettings = azureBakeHandler.findVirtualizationSettings(REGION, bakeRequest)
+    def parameterMap = azureBakeHandler.buildParameterMap(REGION, azureVirtualizationSettings, imageName, bakeRequest, "SomeVersion")
+
+    then:
+    parameterMap.azure_image_name == expectedAzureImageName
+
+    where:
+    buildNumber   || baseName    || imageName                                           || expectedAzureImageName
```
duftler: This should look like:
`buildNumber   | baseName    | imageName                                           || expectedAzureImageName`
for each line.

It is just a convention to separate the givens from the expected. I only learned this recently.

robumberger: Ah, I see. I'll update accordingly.

PR #179, opened 2017-02-10
Allow for multiple RPM repositories

robumberger:
Testing of this change was done in AWS and Azure environments targeting a CentOS 7.1 image. There are 2 ways to specify your target RPM repositories: setting the value of 'yumRepository' in your rosco-local.yml file or via an extended attribute in your bake state. For testing purposes, we used the extended attribute in our bake stage to override the configured value for the 'yumRepository' setting.

To replicate, add an extended attribute named 'repository' in the Bake Configuration of a bake stage. Set the value of this extended attribute to a semi-colon separated string (ie 'http://jfrog-artifactory.repo/artifactory/dotnet-core/;http://nginx.org/packages/centos/7/$basearch/'). In the 'Packages' field in the Bake Configuration, add a package from each repository (ie 'dotnet-core-runtime nginx'). After you start a manual bake, review the bakery log and you should see both packages being installed from their respective source.

### DIFF
--- a/rosco-web/config/packer/install_packages.sh
+++ b/rosco-web/config/packer/install_packages.sh
@@ -55,14 +55,17 @@ function provision_deb() {
 
 function provision_rpm() {
   if [[ "$repository" != "" ]]; then
-    cat > /tmp/spinnaker.repo <<EOF
-[spinnaker]
-name=spinnaker
-baseurl=$repository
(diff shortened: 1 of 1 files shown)


duftler:
Can you please add a short note to the PR description describing how the change was tested?

PR #180, opened 2017-02-10
(core) Don't fail worker threads on top-level exceptions.

duftler:


duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/1421.

duftler:
@lwander please review.

duftler:
Recommend using the `?w=1` magic to eliminate all of the whitespace-only changes from the diff view.

PR #186, opened 2017-02-28
Deprecate google.gce.bakeryDefaults in favor of google.bakeryDefaults

lwander:
This makes the google config more similar to that of the other providers.

The config in rosco-web was left untouched, since it would override existing configurations of rosco with this change in place.


duftler:
Unit tests failed.

lwander:
@duftler good to merge? I reordered the props in the bake handler.

PR #189, opened 2017-03-02
feat(install): Install packer when rosco is installed.

jtk54:


jtk54:
I'm going to add a check if `/packer` exists, should I check the version and upgrade as well?

rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -11,4 +11,13 @@ if [ -z `getent passwd spinnaker` ]; then
   useradd --gid spinnaker spinnaker -m --home-dir /home/spinnaker
 fi
 
+# install packer
+pushd
+mkdir /packer && \
```
lwander: I don't think `/packer` is the right way to do this. I'd create a tempdir and install there, removing the tempdir after.

rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -11,4 +11,13 @@ if [ -z `getent passwd spinnaker` ]; then
...
+  cd /packer \
+  wget https://releases.hashicorp.com/packer/0.12.1/packer_0.12.1_linux_amd64.zip && \
```
lwander: Pull this version into a top-level env var so it's easy to update.

rosco-web/pkg_scripts/postInstall.sh
jtk54: The `Dockerfile` installs to `/packer` and then appends this to `ENV`.

lwander: Dockerfiles are a bit of a special case, since it's more OK to pollute the `/` directory in a containerized environment. This should end up in `/usr/bin`.

jtk54:
Updated & tested.

PR #190, opened 2017-03-03
fix(install): Wrap packer failure in function.

jtk54:
Apparently `set -e` was called somewhere in the install scripts, so if `packer` isn't installed, `packer --version` fails, causing the install script to exit. Wrapping this in a function will let me inspect the return code and decide what to do safely.

### DIFF
--- a/rosco-web/pkg_scripts/postInstall.sh
+++ b/rosco-web/pkg_scripts/postInstall.sh
@@ -12,17 +12,20 @@ if [ -z `getent passwd spinnaker` ]; then
 fi
 
 # install packer
-PACKER_VERSION="0.12.1"
-packer_version=$(packer --version)
-packer_status=$?
-if [ $packer_status -ne 0 ] || [ "$packer_version" -ne "$PACKER_VERSION" ]; then
-  pushd .
-  cd /usr/bin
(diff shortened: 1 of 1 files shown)


rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -12,17 +12,20 @@ if [ -z `getent passwd spinnaker` ]; then
...
-  popd
-fi
+function install_packer() {
+  PACKER_VERSION="0.12.1"
+  local packer_version=$(packer --version)
+  local packer_status=$?
+  if [[ $packer_status -ne 0 ]] || [[ "$packer_version" != "$PACKER_VERSION" ]]; then
+    pushd .
+    cd /usr/bin
+    wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip
```
duftler: Would it make sense to download this file into a directory other than `/usr/bin`? Even though we're deleting it soon after, feels a little off.

duftler: We use a temp dir here: https://github.com/spinnaker/spinnaker/blob/master/InstallSpinnaker.sh#L701

jtk54: It would. I can open another PR because I merged too fast.

PR #191, opened 2017-03-03
fix(install): Use tempdir instead of /usr/bin.

jtk54:


rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -14,16 +14,17 @@ fi
...
     pushd .
-    cd /usr/bin
+    cd $TEMPDIR
     wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip
     apt-get install unzip -y
-    unzip -o "packer_${PACKER_VERSION}_linux_amd64.zip"
-    rm "packer_${PACKER_VERSION}_linux_amd64.zip"
+    unzip -o "packer_${PACKER_VERSION}_linux_amd64.zip" -d /usr/bin
     popd
+    rm -rf $TEMPDIR
```
lwander: This should be outside the `if`, so the $TEMPDIR is deleted even when packer is already installed. (Or move the `mktemp` call into the if statement).

jtk54: I moved the `$TEMPDIR` inside.

PR #198, opened 2017-04-06
fix(Dockerfile): Add all directories below /config.

jtk54:


Dockerfile.slim
```diff
@@ -4,6 +4,8 @@ MAINTAINER delivery-engineering@netflix.com
 
 COPY ./rosco-web/build/install/rosco /opt/rosco
 COPY ./rosco-web/config /opt/rosco
```
duftler: Can't you copy the full directory contents by adding a trailing slash to `config`?

PR #200, opened 2017-04-07
fix(Dockerfile): Navigate Docker's wonky COPY command.

jtk54:
I tested this locally on a Docker container and verified that
the config directory structure looks how we actually want it to this
time. @duftler FYI.  Sorry for the repeated attempts to solve this.


jtk54:
`COPY <from> <to>`

Apparently when `to` is a directory that doesn't exist yet on a container, `COPY` makes a directory when `from` is a directory. When `to` does exist, `COPY` copies the files in `from` into `to` and doesn't actually create a `from` directory in `to`. That's a least my understanding of the way this thing behaves. Not consistent with linux `cp`.

PR #201, opened 2017-05-01
fix(Rosco/Chocolatey) Properly install Chocolatey (NUPKG) packages that include the package version (#1616)

robumberger:
Updated Powershell script that is responsible for installing Chocolatey packages when baking Windows images to properly parse package names. The logic is now the same logic that is used in Rosco when parsing package names.

Package names like 'nodejs.install.7.9.0' are now split into name ('nodejs.install') and version ('7.9.0') and the version is now included when installing the package via Chocolatey command line.


robumberger:
@duftler, @rguthriemsft PTAL

PR #204, opened 2017-05-24
chore(config): Switch to non-deprecated config & remove 12.04 from gce images

lwander:
@duftler because the images between `google.gce.bakeryDefaults.baseImages` and `google.bakeryDefaults.baseImages` are appended to one another, this shouldn't cause any images to be deleted -- unless -- if people have switched to the non-deprecated config (`google.bakeryDefaults.*`), and are depending on one of the three images listed in  `google.gce.bakeryDefaults.baseImages`, they will be overwritten because of how spring merges profiles. I'll check with our bigger users to see if this will affect them.

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -234,59 +234,50 @@ google:
   # with SPINNAKER_GOOGLE_ENABLED or to explicitly set the property's value
   # to true.
   enabled: ${GOOGLE_ENABLED:false}
-  gce:
-    bakeryDefaults:
-      zone: us-central1-f
-      network: default
-      useInternalIp: false
-      templateFile: gce.json
-      baseImages:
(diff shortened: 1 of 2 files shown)


rosco-web/config/rosco.yml
```diff
@@ -234,59 +234,50 @@ google:
...
+    templateFile: gce.json
+    baseImages:
+    # Either sourceImage or sourceImageFamily should be set. If both are set, sourceImage will take precedence.
+    # If you specify isImageFamily, deck will annotate the popup tooltip to indicate that the selected option
+    # represents an image family.
+    #
+    # This is an example of configuring a source image family:
+    # - baseImage:
+    #     id: precise
+    #     shortDescription: v12.04
```
duftler: Should probably update the sample config to note refer to 12.04 anymore either.

PR #212, opened 2017-06-30
feat(google/xpn): Add support for shared VPC networks.

duftler:


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/config/RoscoGoogleConfiguration.groovy
```diff
@@ -60,6 +60,7 @@ class RoscoGoogleConfiguration {
   static class GCEBakeryDefaults {
     String zone
     String network
+    String networkProjectId
```
lwander: Same question here about `Id` as in Halyard

rosco-web/config/packer/gce.json
```diff
@@ -23,6 +24,7 @@
     "account_file": "{{user `gce_account_file`}}",
```
lwander: Oh I see you use `project_id` for the hashicorp vars

duftler:
This just configures the defaults. All attributes can be fully-specified on a given bake request (e.g. vpc network and account); if they are omitted, the defaults are used. Typical pattern is to use the json var file to specify everything and then resolve the correct file via an expression.

PR #215, opened 2017-07-27
inconsistent with BakeRequest.groovy

ekapratama93:
com.netflix.spinnaker.rosco.api.BakeRequest.CloudProviderType use gce and I think its better than using google

### DIFF
--- a/halconfig/images.yml
+++ b/halconfig/images.yml
@@ -147,7 +147,7 @@ docker:
       virtualizationSettings:
         sourceImage: ubuntu:trusty
 
-google:
+gce:
   bakeryDefaults:
     baseImages:
     - baseImage:


duftler:
Those values are deserialized into `RoscoGoogleConfiguration`. The nesting under `gce` has been deprecated: https://github.com/spinnaker/rosco/blob/master/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/config/RoscoGoogleConfiguration.groovy#L46-L56

The values defined in `com.netflix.spinnaker.rosco.api.BakeRequest.CloudProviderType` are referenced by bake requests to indicate target platforms, and `GCE` is in fact used there.

PR #222, opened 2017-10-12
fix(rosco) Make baked artifacts adhere to the proposed standard

brujoand:
This should make the artifacts created in Rosco follow the same standard as those created in Orca (https://github.com/spinnaker/orca/pull/1441)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Artifact.groovy
```diff
@@ -28,4 +28,7 @@ class Artifact {
   String version
   String reference
   Map<String, String> metadata
+  String auth
```
lwander: This should be using the kork-artifacts package instead of recreating this here

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/CloudProviderBakeHandler.groovy
```diff
@@ -110,16 +110,17 @@ abstract class CloudProviderBakeHandler {
    * Returns a decorated artifact for a given bake. Right now this is a generic approach
    * but it could be useful to override this method for specific providers.
    */
-  def Artifact produceArtifactDecorationFrom(BakeRequest bakeRequest, BakeRecipe bakeRecipe, Bake bakeDetails, String cloudProvider) {
+  def Artifact produceArtifactDecorationFrom(BakeRequest bakeRequest, BakeRecipe bakeRecipe, Bake bakeDetails, String cloudProvider, String region) {
     Artifact bakedArtifact = new Artifact(
       name: bakeRecipe?.name,
       version: bakeRecipe?.version,
-      type: cloudProvider,
+      type: "${cloudProvider}/image/${region}",
```
lwander: We're relying on the `location` field in the artifact definition for region

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/Artifact.groovy
brujoand: Cool, I hadn't seen the `kork-artifacts`. Fixing

brujoand:
Now using the `kork-artifacts` for increased DRYness.

brujoand:
@lwander anything else we need to merge this?

PR #225, opened 2017-11-02
fix(rosco) Fix aws enhanced_networking backwards incompatibilities

ikemonn:
Fix packer template for this issue.
https://github.com/spinnaker/spinnaker/issues/2065

There are backwards incompatibilities of enhanced_networking from Packer 1.1.0.

> builder/amazon: Split enhanced_networking into sriov_support and ena_support to support finer grained control. Use packer fix <template.json> to automatically update your template to use ena_support where previously there was only enhanced_networking. Make sure to also add sriov_support if you need that feature, and to ensure ena_support is what you intended to be in your template. [GH-5284]

https://github.com/hashicorp/packer/blob/master/CHANGELOG.md#110-september-12-2017


costimuraru:
👍

duftler:
Thanks for the fix. There is one other related change that I believe is required:

Doesn't this line need to now reference the new user variable? https://github.com/spinnaker/rosco/blob/master/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy#L127

ikemonn:
@duftler 
Thank you for your review.

> Doesn't this line need to now reference the new user variable? 

Yes.
`buildPackerCommand` makes args for packer to use user variable. 
So this line is needed to change to new user variable name.

https://github.com/spinnaker/rosco/blob/8f0564784f6a04631b7531c95cd876a93e4503d2/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy#L27-L34

Please take another look.

duftler:
Needs an additional rebase.

duftler:
Please rebase at head instead of merging master.

ikemonn:
@duftler 
Thanks.
I created new PR and close this one.
Please check it.

PR #226, opened 2017-11-02
fix(rosco): Revert Make pause_before configurable (#224)

tpulmano:
This reverts commit 8f0564784f6a04631b7531c95cd876a93e4503d2.

Per this Packer issue `pause_before` isn't being expanded: https://github.com/hashicorp/packer/issues/5115

### DIFF
--- a/rosco-web/config/packer/aws-ebs.json
+++ b/rosco-web/config/packer/aws-ebs.json
@@ -17,8 +17,7 @@
     "package_type": "",
     "packages": "",
     "upgrade": "",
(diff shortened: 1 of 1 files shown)


avram:
This was my mistake. We identified that we wanted this configurable and I eagerly PR'ed the obvious "quick fix" -- but Packer doesn't let us template that value.

duftler:
LGTM.
I suppose you could add an additional `shell` provisioner just before and have it's referenced script do nothing but pause for some parameterized amount of time.

PR #231, opened 2018-01-04
updated ami ids for baseImage

LeSandeep:
We're updating the trusty base image ami id's to a newer version.

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -101,51 +101,53 @@ aws:
         shortDescription: v14.04
         detailedDescription: Ubuntu Trusty Tahr v14.04
         packageType: deb
+        # The following AMI ID's were retrieved from here:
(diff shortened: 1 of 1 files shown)


rosco-web/config/rosco.yml
```diff
@@ -105,47 +105,47 @@ aws:
       - region: us-east-1
         virtualizationType: hvm
         instanceType: t2.micro
-        sourceAmi: ami-9eaa1cf6
+        sourceAmi: ami-9d751ee7
```
lwander: Could you link to where you got these as comments for the various ami IDs? This will affect all rosco users who are depending on the default configuration & it would be nice for them to easily verify (in the config) what they're using as a base image.

LeSandeep: Sure thing. Update the diff with the link.

PR #232, opened 2018-01-11
feat(bake) Added AWS spot pricing configuration when baking AMI's. Defaults to 'auto'.

robumberger:
Packer defaults to using on-demand pricing when creating AWS AMI's. Given that most bakes take considerably less than an hour, using on-demand pricing is wasteful since you pay for the full hour. This PR enables Packer to use spot pricing and defaults to 'auto' which allows Packer to automatically discover the best spot price. Setting 'spotPrice' to '0' (zero) causes Packer to use on-demand pricing.

### DIFF
--- a/halconfig/images.yml
+++ b/halconfig/images.yml
@@ -13,26 +13,36 @@ aws:
         instanceType: t2.micro
         sourceAmi: ami-d4aed0bc
         sshUserName: ubuntu
+        spotPrice: auto
+        spotPriceAutoProduct: Linux/UNIX (Amazon VPC)
       - region: us-west-1
         virtualizationType: hvm
         instanceType: t2.micro
         sourceAmi: ami-4f285a2f
         sshUserName: ubuntu
+        spotPrice: auto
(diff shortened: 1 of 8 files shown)


robumberger:
@duftler PTAL - thanks

halconfig/images.yml
```diff
@@ -13,26 +13,36 @@ aws:
         instanceType: t2.micro
         sourceAmi: ami-d4aed0bc
         sshUserName: ubuntu
+        spotPrice: auto
+        spotPriceAutoProductType: Linux/UNIX (Amazon VPC)
```
duftler: This is `spotPriceAutoProduct` everywhere else.

duftler:
The code LGTM, but I lack some intuition about how spot pricing works to judge whether the change makes sense overall (seems logical, but I don't have lots of context here).
@gardleopard Can you PTAL as well?

halconfig/images.yml
robumberger: Fixed. Thanks for the catch.

PR #235, opened 2018-02-09
feat(bake) The addition of spot changed the defaults, swap them back

mikerowehl:
A change was made to rosco recently to add the ability to use spot instances when baking. In the process it also changed spot to be the default. The functionality itself is fantastic, but defaulting to spot actually comes along with a few additional requirements. That causes bakes to start failing on update. So defaulting to the same bake style as we had in 0.102.0 and before, but leaving all the options there so users can turn spot on.

- Change the spotPrice default in the virtualizationSettings to 0 instead of auto. Set to auto it was defaulting to doing bakes using spot, which is a big change from the way we have been operating.
- Need to quote the spot price in the templates, otherwise packer throws an error with "expected type 'string', got unconvertible type 'float64'"
- Leave the spotPriceAutoProduct set in the virtualization settings, so users can set just aws_spot_price in their bake stage and have it work as expected.

### DIFF
--- a/halconfig/images.yml
+++ b/halconfig/images.yml
@@ -13,35 +13,35 @@ aws:
         instanceType: t2.micro
         sourceAmi: ami-d4aed0bc
         sshUserName: ubuntu
-        spotPrice: auto
+        spotPrice: 0
         spotPriceAutoProduct: Linux/UNIX (Amazon VPC)
       - region: us-west-1
         virtualizationType: hvm
         instanceType: t2.micro
         sourceAmi: ami-4f285a2f
         sshUserName: ubuntu
-        spotPrice: auto
+        spotPrice: 0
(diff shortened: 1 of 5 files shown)


duftler:
Seems reasonable to me.

@robumberger Can you please take a look as well? It does seem like the default should have remained the same, and the new behavior enabled as an option.

robumberger:
LGTM. We just changed our bake workflow and ran into a related issue. Due to the way spot works, if you have ENA enabled ("aws_ena_support"), this caused bakes to fail since ENA is applied after the EC2 instance is stopped. With spot enabled, stopping the instance causes the underlying image to no longer available, so when Packer tries to set the ENA flag, it fails.

duftler:
Thanks @robumberger . So do you think adding a comment (warning not to enable both spot and ENA) to the checked-in templates makes sense?

PR #238, opened 2018-03-29
fix(bake): Fix broken AWS Windows bake script

brantburnett:
The current aws-windows-2012-r2.json Packer script is malformed, causing
all bakes of AWS Windows AMIs to fail if they are using the default bake
script.  https://github.com/spinnaker/spinnaker/issues/2598


brantburnett:
@spinnaker/reviewers

Just checking in on this pull req since it's been more than the 3 days mentioned in the instructions.

Thanks!

PR #241, opened 2018-04-05
feat(artifacts): Update GCE bake to send image reference

ezimanyi:
When baking a GCE image, set the reference of the produced artifact to the URI for the produced image

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
@@ -174,4 +174,18 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
     return managedGoogleAccount
   }
 
+  @Override
+  def String getArtifactReference(BakeRequest bakeRequest, Bake bakeDetails) {
+    RoscoGoogleConfiguration.ManagedGoogleAccount managedGoogleAccount = resolveAccount(bakeRequest)
+    def project = managedGoogleAccount.getProject()
+    def imageName = bakeDetails.image_name
+
+    // TODO(ezimanyi): Remove hard-coding of image URI
+    // Either get packer to directly return the generated URI (preferred) or send a request to
+    // clouddriver to convert the project/image combination into a URI using the Google compute API
+    def uri = "https://www.googleapis.com/compute/v1/projects/$project/global/images/$imageName"
+    return uri
+  }
+
+
 }
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
@@ -175,7 +175,7 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
   }
 
   @Override
-  def String getArtifactReference(BakeRequest bakeRequest, Bake bakeDetails) {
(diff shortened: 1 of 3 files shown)


ezimanyi:
@gardleopard : Do you mind taking a look at this proposed change to artifact decoration in Rosco?

My goal here is to have the artifacts produced by Rosco follow the standard artifact format.  There are thus two changes I'm proposing:
-Currently `version` is set to information about the provenance of the artifact (ex: the Jenkins build that produced it, etc.) rather than to a `version` of the artifact itself.  For now I'm clearing the field as a `gce/image` is not versions (and I don't believe an AMI is either) so I don't think this field is relevant.  We could move this to `metadata` or `provenance` if this is still useful (and not redundant with the other information in `metadata`
-For GCE the `reference` is currently set to the name of the image; I'm changing this to the URI as this field should contain a unique identifier for the artifact

Can you let me know if you think these changes make sense, and if there are any downstream effects you can think of that I may not be aware of?  Thanks!

ezimanyi:
@gardleopard : Just pinging this again quickly to see if you minded checking whether you see any issues with this change to bake artifact decoration.  Thanks a lot!  -Eric

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
```diff
@@ -174,4 +174,18 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
...
+    def project = managedGoogleAccount.getProject()
+    def imageName = bakeDetails.image_name
+
+    // TODO(ezimanyi): Remove hard-coding of image URI
+    // Either get packer to directly return the generated URI (preferred) or send a request to
+    // clouddriver to convert the project/image combination into a URI using the Google compute API
+    def uri = "https://www.googleapis.com/compute/v1/projects/$project/global/images/$imageName"
+    return uri
+  }
+
```
lwander: Extra newline?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
```diff
@@ -174,4 +174,18 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
...
+  def String getArtifactReference(BakeRequest bakeRequest, Bake bakeDetails) {
+    RoscoGoogleConfiguration.ManagedGoogleAccount managedGoogleAccount = resolveAccount(bakeRequest)
+    def project = managedGoogleAccount.getProject()
+    def imageName = bakeDetails.image_name
+
+    // TODO(ezimanyi): Remove hard-coding of image URI
+    // Either get packer to directly return the generated URI (preferred) or send a request to
+    // clouddriver to convert the project/image combination into a URI using the Google compute API
+    def uri = "https://www.googleapis.com/compute/v1/projects/$project/global/images/$imageName"
+    return uri
```
lwander: Can you inline this `def ... return`?

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
```diff
@@ -174,4 +174,18 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
     return managedGoogleAccount
   }
 
-}
\ No newline at end of file
+  @Override
+  def String getArtifactReference(BakeRequest bakeRequest, Bake bakeDetails) {
```
lwander: I think this `def` isn't needed, just `String <funcName>(...)` is enough

PR #242, opened 2018-04-06
chore(docker): update packer

lwander:


danielpeach:
Should the other Dockerfile be updated too?

lwander:
We don't maintain the other dockerfile -- so possibly? I guess this change is benign enough but in general we leave those untouched

PR #243, opened 2018-04-06
feat(bake/manifests): Rudimentary bake manifest support

lwander:
Still missing:

1. Downloading helm values files
2. Setting helm --set values
3. Deleting staged helm charts
4. Creating non base64 artifacts
5. Polling of running bake tasks -- the bake is currently synchronous

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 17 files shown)


lwander:

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -0,0 +1,35 @@
+package com.netflix.spinnaker.rosco.services;
```
ezimanyi: Do we need the standard copyright text at the top of this and other new classes?

lwander: Good catch, looks like my intellij hasn't been configured for this project yet

lwander:
It's not really backwards compatible, at least from a url path perspective. 

The prior url paths were 

`/api/v1/{region}/bake`

Manifests don't have any region associated with them -- sticking a placeholder value in there is hacky and registering another path that uses a different `/api/v1/<something else>...` has potential to collide with the original path in the future.

Ideally, this probably should have been registered as

`/api/v1/bake/{region}` or something similar, giving other endpoints the chance to register something other than `{region}` as the first path token after `/api/v1`. 

Starting with `v2`, we have `/api/v2/manifest/bake`, and can add `/api/v2/image/bake/region/{region}`

PR #244, opened 2018-04-10
chore(packer) - upgrade packer binary to 1.2.2

imosquera:
Upgrades packer binary to 1.2.2 in both Docker and Docker.slim.

This PR addresses this issue here:
https://github.com/spinnaker/spinnaker/issues/2545


lwander:
LGTM

PR #245, opened 2018-04-12
feat(core): add shutdown hook to avoid killing running jobs

lwander:
Given that the v2 deployer doesn't wait for rosco to drain all work before shutting down, this gives Kubernetes a chance (with `terminationGracePeriodSeconds`) to allow rosco to finish any ongoing bakes.


lwander:
https://github.com/spinnaker/spinnaker/issues/2288

PR #246, opened 2018-04-13
fix(bake/manifests): delete templates when request finishes

lwander:

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
@@ -5,6 +5,9 @@ import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import com.netflix.spinnaker.kork.core.RetrySupport;
 import com.netflix.spinnaker.rosco.jobs.BakeRecipe;
 import com.netflix.spinnaker.rosco.services.ClouddriverService;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FileUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import retrofit.client.Response;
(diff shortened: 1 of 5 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
```diff
@@ -47,4 +52,39 @@ public Path downloadArtifactToTmpFile(Artifact artifact) throws IOException {
 
     return path;
   }
+
+  public static class BakeManifestEnvironment {
+    @Getter
+    final private Path stagingPath = Paths.get(System.getProperty("java.io.tmpdir"), UUID.randomUUID().toString());
+
+    public BakeManifestEnvironment() {
+      stagingPath.toFile().mkdirs();
```
duftler: Might be worth a `log.debug` here.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
```diff
@@ -47,4 +52,39 @@ public Path downloadArtifactToTmpFile(Artifact artifact) throws IOException {
...
+    }
+
+    private boolean recursiveDeleteFile(File file) {
```
duftler: Can use `FileSystemUtils.deleteRecursively()`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
lwander: :+1:

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
lwander: Wasn't imported yet -- will do

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/TemplateUtils.java
lwander: Upgraded to `warn` on failure

duftler: My thinking was if it doesn't fail, but isn't obvious where it was created (or isn't where one expected), they could use debug logging to troubleshoot.

PR #248, opened 2018-04-18
feat(bakery): GCE image bakery supports manifest files

ezimanyi:


ezimanyi:
This change adds a "manifest" post-processor to the default packer template for GCE, and uses the image name in the generated manifest file instead of scraping the logs.  Two things to note:
* This change is backwards-compatible; if not manifest file is found (ie, an older packer template was used) then we fall back to the same log scraping as was done before
* Users using a custom can get this new functionality by adding a manifest post-processor to their template with an output set to the user variable `manifestFile`

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/google/GCEBakeHandler.groovy
```diff
@@ -150,11 +157,18 @@ public class GCEBakeHandler extends CloudProviderBakeHandler {
   Bake scrapeCompletedBakeResults(String region, String bakeId, String logsContent) {
     String imageName
 
-    // TODO(duftler): Presently scraping the logs for the image name. Would be better to not be reliant on the log
-    // format not changing. Resolve this by storing bake details in redis.
-    logsContent.eachLine { String line ->
-      if (line =~ IMAGE_NAME_TOKEN) {
-        imageName = line.split(" ").last()
+    if (packerManifestController.manifestExists(bakeId)) {
+      log.info("Using manifest file to determine baked artifact")
```
lwander: I might throw the bakeId in this log line to help with tracebility.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackerManifestController.java
```diff
@@ -0,0 +1,63 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.rosco.providers.util.PackerManifest.PackerBuild;
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+
+public class PackerManifestController {
```
lwander: Usually we reserve the word "controller" for web endpoints, e.g. spring `@RestController`s. This might best be called a "service"?

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/util/PackerManifestControllerSpec.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+package com.netflix.spinnaker.rosco.providers.util
+
+import spock.lang.Specification
+
+import java.nio.file.Paths
+
+class PackerManifestControllerSpec extends Specification implements TestDefaults {
+    def "getManifestFileName"() {
+        setup:
```
lwander: Indentation looks a little funny here

PR #250, opened 2018-04-21
Upgrade to Gradle 4.7

jkschneider:
Taking advantage of spinnaker-gradle-project [4.0.0](https://github.com/spinnaker/spinnaker-gradle-project/pull/38).

See related PRs in:
* [Clouddriver](https://github.com/spinnaker/clouddriver/pull/2545)
* [Deck](https://github.com/spinnaker/deck/pull/5220)
* [Echo](https://github.com/spinnaker/echo/pull/257)
* [Fiat](https://github.com/spinnaker/fiat/pull/223)
* [Gate](https://github.com/spinnaker/gate/pull/530)
* [Orca](https://github.com/spinnaker/orca/pull/2170)
* [Front50](https://github.com/spinnaker/front50/pull/310)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/docker/DockerImageNameFactory.groovy
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
duftler: Could verify this by viewing history for each file, but fairly certain the copyright for all of these *Util classes should be Schibsted.

jkschneider: Perhaps at some point copyrights could be relegated to "Spinnaker Contributors" or something like that. Believe there are other projects that do something similar.

jkschneider: I'll just remove license updates here. It was just an attempt to get nagging warnings out of the build anyway.

PR #251, opened 2018-04-23
feat(bake/manifest): allow multiple input artifacts

lwander:


lwander:
That's right. The issue is different template renderers will handle this differently, and for the next two we plan to support (jinja, jsonnet), all that really makes sense is to pass in multiple files in a particular order, there is no special "first" artifact. I personally lean toward having all share the same stage/operation specification to make handling this easier between services & template renderers.

ezimanyi:
OK, makes sense, particularly if it's a quirk of this particular template renderer that the first element is "special" and not something we'll generally have to deal with...I just generally get nervous when I see arrays where some elements are "special" but looks like it might be the best approach here.

PR #252, opened 2018-04-23
feat(bake/artifacts): Pass artifacts to packer

ezimanyi:
Updates the rosco bake endpoint to accept a list of artifacts and passes these artifacts to packer.  Packer does not yet consume these artifacts, which will be implemented in a later pull request.


ezimanyi:
This PR just handles the plumbing so that Rosco can accept a list of artifacts and pass these to the packer builder as a JSON file.  I've deferred the update to `install_packages.sh` to the next PR to keep things reasonably small in scope (though I did just tiny change to `install_packages.sh` to remove clean up the uploaded file).

PR #253, opened 2018-04-24
feat(bake/artifacts): Pass artifacts to packer

ezimanyi:
Updates the rosco bake endpoint to accept a list of artifacts and passes these artifacts to packer.  Packer does not yet consume these artifacts, which will be implemented in a later pull request.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackerArtifactService.java
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackerArtifactService.java
@@ -32,7 +32,13 @@ public class PackerArtifactService {
   private Path tempDir = Paths.get(System.getProperty("java.io.tmpdir"));
   private ObjectMapper objectMapper = new ObjectMapper();
 
-  public String writeArtifactsToFile(String bakeId, List<Artifact> artifacts) {
+  public PackerArtifactService() throws IOException {
+    if (!Files.isDirectory(tempDir)) {
+      Files.createDirectories(tempDir);
+    }
+  }
+
+  public Path writeArtifactsToFile(String bakeId, List<Artifact> artifacts) {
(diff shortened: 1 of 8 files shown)


ezimanyi:
This PR just handles the plumbing so that Rosco can accept a list of artifacts and pass these to the packer builder as a JSON file. I've deferred the update to install_packages.sh to the next PR to keep things reasonably small in scope (though I did just tiny change to install_packages.sh to clean up the uploaded file).

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackerArtifactService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PackerArtifactService {
+
+  private Path tempDir = Paths.get(System.getProperty("java.io.tmpdir"));
+  private ObjectMapper objectMapper = new ObjectMapper();
+
+  public String writeArtifactsToFile(String bakeId, List<Artifact> artifacts) {
```
lwander: Maybe returning a `Path` will make this function a little clearer

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackerArtifactService.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.util.ArrayList;
+import java.util.List;
+
+public class PackerArtifactService {
+
+  private Path tempDir = Paths.get(System.getProperty("java.io.tmpdir"));
```
lwander: I might just be paranoid, but I'd check that this exists (& try to create it if not)

PR #257, opened 2018-05-02
feat(artifacts): Enable rosco to install packages passed as artifacts

ezimanyi:


rosco-web/config/packer/install_packages.sh
```diff
@@ -16,6 +19,49 @@ repository=`echo $repository | sed 's/^"\(.*\)"$/\1/'`
...
+    sudo apt-get update
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install --force-yes -y jq
+  fi
+
+  if [[ "$package_type" == "rpm" ]]; then
+    sudo yum -y install jq
+  fi
+}
+
+function remove_jq() {
```
lwander: What if jq was already installed before this script ran?

ezimanyi: Good point---it's probably unlikely it would be on a newly-provisioned image, but I shouldn't assume this.  It should be easy to check this before doing the install/uninstall.

PR #260, opened 2018-05-15
fix(manifest): Fix using base64 artifacts in manifest bakery

posox:
Use MD5 hash in tmp path instead reference

spinnaker/spinnaker#2781


posox:
@lwander commit message updated

PR #261, opened 2018-05-16
feat(bake/manifests): Add namespace support for Helm bakery

posox:
This patch adds support passing namespace for Helm bakery.

link: spinnaker/spinnaker#2763

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
@@ -15,6 +15,7 @@ public class BakeManifestRequest {
   String outputName;
   List<Artifact> inputArtifacts;
   Map<String, Object> overrides;
+  String namespace;
 
   enum TemplateRenderer {
     HELM2;
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
@@ -15,7 +15,7 @@ public class BakeManifestRequest {
   String outputName;
   List<Artifact> inputArtifacts;
   Map<String, Object> overrides;
-  String namespace;
+  Map<String, Object> properties;
 
   enum TemplateRenderer {
     HELM2;
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
@@ -2,7 +2,6 @@ package com.netflix.spinnaker.rosco.manifests;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
-import com.netflix.spinnaker.rosco.jobs.BakeRecipe;
 import lombok.Data;
 
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
```diff
@@ -15,6 +15,7 @@
   String outputName;
   List<Artifact> inputArtifacts;
   Map<String, Object> overrides;
+  String namespace;
```
lwander: This is very specific to helm & k8s, I'm not sure this is the right spot. Is this the only way namespace can be configured?

ethanfrogers: Is there anyway we could have a helm specific bake manifest request? Having this property exposed makes it easier to being existing Helm charts in because it's common practice to use things like `{{ .Release.Name }}` and `{{ .Release.Namespace }}`.

lwander: Yes that sounds best -- @posox can you factor this into a separate helm bake manifest request class?

posox: @lwander @ethanfrogers I would prefer to use common property like `properties` with type `Map<String, Object>` or something else.
As I correct understand new request class requires new api call like `/api/v2/manifest/bake/helm`.

lwander: We could have a new request class, or just pass the type as a parameter to the generic bake manifest endpoint

posox: @lwander added `properties` parameter to the generic request

lwander: Awesome

> As I correct understand I should add new RequestMapping like /api/v2/manifest/bake/helm.

I think that's up to you -- we can either have a `type` field that decides how to cast the incoming request, or have different endpoints that orca decides call.

ethanfrogers: hey @posox have you made any progress on refactoring this?

posox: @ethanfrogers yes, I'm working on refactoring.
@lwander I have a question. I don't understand how we can use `type` field in request for passing namespace.
As i correct understand it can be done using new field in request or using new endpoint for Helm bakery.

ethanfrogers: @posox i think @lwander was saying use a `type` field to denote the type of BakeManifestRequest (Helm, other) and then cast the request to a concrete `HelmBakeManifestRequest`. That way we can support other types in the future and put implementation specific fields (`namespace` for Helm) on the implementation specific classes. Does that make sense?

posox: @ethanfrogers  how I can cast `BakeManifestRequest` to `HelmBakeManifestRequest`? In `BakeManifestRequest` the `namespace` field does not exists, so request cannot be casted correctly. I can change `RequestBody` to `HelmBakeManifestRequest`, but it looks strange.

posox:
Please review: @spinnaker/reviewers

ethanfrogers:
ping @lwander

posox:
@lwander it should be merged with https://github.com/spinnaker/orca/pull/2241

posox:
@spinnaker/reviewers please merge

lwander:
Thank you @posox!

PR #265, opened 2018-05-24
fix(artifacts): Guard access to artifacts file in bake script

ezimanyi:


ezimanyi:
Fixes https://github.com/spinnaker/spinnaker/issues/2835

PR #269, opened 2018-07-23
fix(bake/manifests): capture onny stdout in Helm bakery

posox:
issue: https://github.com/spinnaker/spinnaker/issues/3063

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeStatus.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeStatus.groovy
@@ -53,6 +53,9 @@ class BakeStatus implements Serializable {
   String logsContent
 
   @JsonIgnore
+  String stdOut
+
+  @JsonIgnore
   long createdTimestamp
 
   @JsonIgnore
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
@@ -59,8 +59,9 @@ class JobExecutorLocal implements JobExecutor {
       new Action0() {
         @Override
         public void call() {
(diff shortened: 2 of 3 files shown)


ethanfrogers:
@posox do you think it's worth capturing `stderr` as well so we can communicate any errors/warnings to the user?

ethanfrogers:
I guess that's a fair point. however, in that case, warnings won't ever get caught which may lead to some difficult to debug and unexpected behavior.

posox:
@ethanfrogers I can add more logs in Rosco.

ethanfrogers:
hopefully @lwander can confirm this, but i'm a bit concerned that separating out both log streams may lead to some unintended issues on the packer side of rosco. for example, lets say that packer logs to both `stdout` and `stderr` during a bake. a combined stream would have those logs in the correct order where a separate stream, combined after the fact, would show them out of order.

i haven't tested this theory but it seems like that might be the case from looking at the code.

lwander:
This LGTM

@ethanfrogers we still capture both in `stdOutAndErr` right? The tricky thing with output streams is that you can't guarantee that the output will be in the correct order, even when printing directly to a terminal. 

I agree that we should be surfacing `stdOutAndErr` in the case where the bake fails, though.

posox:
> we still capture both in stdOutAndErr right?

that's right

Please review / merge: @spinnaker/reviewers

posox:
@spinnaker/reviewers: please review / merge

duftler:
@lwander Can you please have a look at this and merge if ready?

guoyongzhang:
This PR seems to have a negative effect capturing bake job logs.  In my bake job (regular bake, not bake manifest) run, only the first 2 lines were captured.  The packer did complete the job, but since logs were lost, the stage finished with TERMINAL status.
Once I revert this commit in rosco, I can see the full logs were captured running the exact same bake job, and the stage finished with SUCCEEDED status.

ethanfrogers:
ouch, i was worried about this. @posox @lwander i wonder if we should revert this and figure out a better solution.

lwander:
Let's revert before 1.10 is released

PR #275, opened 2018-08-31
feat(bake/manifests): use custom output artifact name in helm bakery

posox:
issue: https://github.com/spinnaker/spinnaker/issues/3108

this patch depends on: https://github.com/spinnaker/orca/pull/2402

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java
@@ -12,6 +12,7 @@ import java.util.Map;
 public class BakeManifestRequest {
   TemplateRenderer templateRenderer;
   String outputName;
+  String outputArtifactName;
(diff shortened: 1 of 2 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestRequest.java
```diff
@@ -8,4 +8,5 @@
 @EqualsAndHashCode(callSuper = true)
 public class HelmBakeManifestRequest extends BakeManifestRequest {
     String namespace;
+    String outputArtifactName;
```
lwander: I think you can put this in `BakeManifestRequest`, since this isn't specific to helm.

posox: @lwander fixed

duftler:
@lwander Can you please have a look at this and merge if ready?

PR #278, opened 2018-09-18
fix(bake/oracle): scrape image id as well as name.

guoyongzhang:
Scrape the oracle bake result for additional image id.


guoyongzhang:
Could someone please review and merge this PR? Thanks.

PR #310, opened 2018-12-20
fix(packer): Fix spot pricing flags

ezimanyi:
Packer 1.3 now enforces that aws_spot_price_auto_product can only be passed if aws_spot_price is set to "auto". The default settings that ship with Spinnaker set aws_spot_price to "0" and aws_spot_price_auto_product non-null. Gate passing through the product value so that it is only passed through if the spot price is "auto".

### DIFF
--- a/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
+++ b/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
@@ -624,40 +624,40 @@ class AWSBakeHandlerSpec extends Specification implements TestDefaults {
 
   void 'sends spot_price_auto_product iff spot_price is set to auto'() {
     setup:
-    def imageNameFactoryMock = Mock(ImageNameFactory)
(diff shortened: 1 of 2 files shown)


ezimanyi:
Fixes https://github.com/spinnaker/spinnaker/issues/3764

Also ping @robumberger who added spot pricing support to take a look.

If this looks good I'll cherry-pick into 1.11

rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy
```diff
@@ -622,6 +622,44 @@ class AWSBakeHandlerSpec extends Specification implements TestDefaults {
       1 * packerCommandFactoryMock.buildPackerCommand("", parameterMap, null, "$configDir/$awsBakeryDefaults.templateFile")
   }
 
+  void 'sends spot_price_auto_product iff spot_price is set to auto'() {
```
duftler: Formatting differs from surrounding test cases.

ezimanyi:
@spinnakerbot cherry-pick 1.11

PR #315, opened 2019-01-11
fix(packer): Determine image name from unencrypted AMI log line

jonsie:
@duftler 

We upgraded to Packer 1.3+ (1.3.3, specifically) and the line with the AMI name changed.  This accounts for that change.  Not entirely thrilled about this, but it works and follows the current pattern.

Merge it if you like, or leave it here just for posterity - either way we're using it in our own deploy.  I think ultimately resolving this issue the way @duftler outlined in the TODO comment is the way to go.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
@@ -29,6 +29,7 @@ import org.springframework.stereotype.Component
 public class AWSBakeHandler extends CloudProviderBakeHandler {
 
   private static final String IMAGE_NAME_TOKEN = "amazon-(chroot|ebs): Creating the AMI:"
+  private static final String UNENCRYPTED_IMAGE_NAME_TOKEN = "(==> |)amazon-(chroot|ebs): Creating unencrypted AMI"
 
   ImageNameFactory imageNameFactory = new ImageNameFactory()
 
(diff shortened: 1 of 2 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -29,6 +29,7 @@ import org.springframework.stereotype.Component
 public class AWSBakeHandler extends CloudProviderBakeHandler {
 
   private static final String IMAGE_NAME_TOKEN = "amazon-(chroot|ebs): Creating the AMI:"
+  private static final String UNECRYPTED_IMAGE_NAME_TOKEN = "(==> |)amazon-(chroot|ebs): Creating unencrypted AMI"
```
duftler: s/UNECRYPTED/UNENCRYPTED

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -161,6 +162,9 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
     logsContent.eachLine { String line ->
       if (line =~ IMAGE_NAME_TOKEN) {
         imageName = line.split(" ").last()
+      } else if (line =~ UNECRYPTED_IMAGE_NAME_TOKEN) {
```
duftler: Maybe add a test to the unit tests here: https://github.com/spinnaker/rosco/blob/master/rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandlerSpec.groovy

PR #327, opened 2019-01-28
feat(provider/azure): Use Azure Managed Disk instead of VHD in bake

michaeljqzq:
* Update json for packer command

* Update parameter map

* Increase the image name length to 75

* Update the logic to fetch the managed image name and id

* Fix test case: AuzreBakeHandlerSpec

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ allprojects {
 
 
   ext {
-    spinnakerDependenciesVersion = '1.17.0'
+    spinnakerDependenciesVersion = '1.19.0'
     if (project.hasProperty('spinnakerDependenciesVersion')) {
       spinnakerDependenciesVersion = project.property('spinnakerDependenciesVersion')
     }


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -61,9 +62,11 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
     // TODO(duftler): Presently scraping the logs for the image name. Would be better to not be reliant on the log
     // format not changing. Resolve this by storing bake details in redis.
     logsContent.eachLine { String line ->
-      if (line =~ IMAGE_NAME_TOKEN) {
-        imageName = line.split("/").last()
-        ami = "https:" + line.split(":").last()
+      if (line.startsWith(IMAGE_NAME_TOKEN)) {
+        imageName = line.substring(IMAGE_NAME_TOKEN.size())
+      }
+      if (line.startsWith(IMAGE_ID_TOKEN)) {
```
mybayern1974: Are there two business scenarios mapped to IMAGE_NAME_TOKEN and IMAGE_ID_TOKEN separately?  If so, suggest adding comments here telling what they are and better with sample values of them.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -61,9 +62,11 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
...
+        ami = line.substring(IMAGE_ID_TOKEN.size())
```
mybayern1974: If there are two business scenarios mapped with these 2 cases, suggest renaming the two variables, at least renaming the "ami" since this is the AWS naming convention.

duftler: Would be nice to rename it to something not aws-specific, but kind of stuck with it for historical reasons; the upstream services rely on that name for this object: https://github.com/spinnaker/orca/blob/master/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/api/Bake.groovy#L36

On the GCE side, we don't use it, but that's only because there isn't a meaningful distinction between id and name for a vm image (well, it's not so much that there isn't a distinction, it's that the id isn't relied on and we address the images solely by name).

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -98,21 +102,28 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
...
-    parameterMap.azure_image_name = azureImageName
+    // Cut the 75 characters for image name due to Azure Managed Disk limitation
+    // The imageName is made up by 4 parts: PACKAGE_NAME-PACKAGE_TYPE-TIMESTAMP-OS
+    // A potential issue is that if the package name is too long(let's say 80), then if we cut the name by 75
+    // the final image name will only contains part of the package name. So it will be the same for every build
+    // A better way to fix is cutting the package name if it is too long and always keep the "PACKAGE_TYPE-TIMESTAMP-OS" part
+    // But since the imageName here is already combination of 4 parts, not easy to figure out what the exact package name is
+    azureImageName = azureImageName.length() <= 75 ? azureImageName : azureImageName.substring(0, 75)
+    
+
```
duftler: Extraneous blank line.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
```diff
@@ -98,21 +102,28 @@ public class AzureBakeHandler extends CloudProviderBakeHandler{
...
     azureImageName = azureImageName.replaceAll(/[\-\.]+$/, "")
-
-    parameterMap.azure_image_name = azureImageName
+    // Cut the 75 characters for image name due to Azure Managed Disk limitation
+    // The imageName is made up by 4 parts: PACKAGE_NAME-PACKAGE_TYPE-TIMESTAMP-OS
+    // A potential issue is that if the package name is too long(let's say 80), then if we cut the name by 75
+    // the final image name will only contains part of the package name. So it will be the same for every build
+    // A better way to fix is cutting the package name if it is too long and always keep the "PACKAGE_TYPE-TIMESTAMP-OS" part
+    // But since the imageName here is already combination of 4 parts, not easy to figure out what the exact package name is
+    azureImageName = azureImageName.length() <= 75 ? azureImageName : azureImageName.substring(0, 75)
```
duftler: This being Groovy, you can replace this line with:
`azureImageName = azureImageName.take(75)`

http://docs.groovy-lang.org/docs/groovy-2.3.2/html/api/org/codehaus/groovy/runtime/StringGroovyMethods.html#take(java.lang.CharSequence,%20int)

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
michaeljqzq: In order not to break anything, I didn't change the naming of variable

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/azure/AzureBakeHandler.groovy
michaeljqzq: fixed

kevinheins:
@spinnakerbot cherry-pick 1.12

PR #342, opened 2019-03-04
feat(manifests): jinja manifest rendering

ethanfrogers:
adds support for jinja as a manifest rendering engine. i would have
preferred to use `jinjava` for the rendering but job execution is really
coupled to a command line call so i opted for `jinja2-cli`.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -20,6 +20,10 @@ RUN wget https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get &&
   ./get && \
   rm get
 
+RUN wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py && \
+    python /tmp/get-pip.py && \
+    pip install jinja2-cli --upgrade
+
 RUN adduser -D -S spinnaker
 
 USER spinnaker
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Armory
+ * Copyright 2019 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


ethanfrogers:
@ezimanyi PTAL.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/jinja/JinjaBakeManifestService.java
```diff
@@ -0,0 +1,89 @@
...
+  JinjaTemplateUtils templateUtils;
+
+  @Override
+  public boolean handles(String type) {
+    return type.equals("jinja");
+  }
+
+  public Artifact bake(Map<String, Object> request) {
+    ObjectMapper mapper = new ObjectMapper();
+    JinjaBakeManifestRequest bakeManifestRequest = mapper.convertValue(request, JinjaBakeManifestRequest.class);
```
ethanfrogers: I'm not `try/catching` this so there's a chance that it could throw an exception. What's a good pattern for handling this here? Throwing an `IllegalArgumentException`?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2019 Netflix, Inc.
```
ezimanyi: s/Netflix/Armory/

Dockerfile.slim
```diff
@@ -20,6 +20,10 @@ RUN wget https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get &&
   ./get && \
   rm get
 
+RUN wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py && \
```
ezimanyi: This looks to install python 2.7 which is EOL in 10 months---could we install something more future-proof?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/jinja/JinjaBakeManifestService.java
ezimanyi: Maybe, though it might also just be reasonable to propagate any generated exception here.

ethanfrogers:
> How much more work would it be to have this run thorough `jinjava`?

I _think_ it'd be a fair amount of work, at least for me, because the the default `JobExecutor` implementation is coupled to shell execution. We'd need to write a different implementation that executes bits of code rather than shelling out. 

> There are security concerns around allowing end-users to execute user-supplied Jinja templates
Totally fair point. I hadn't even considered this. Good example, too. If we decide that the above is worth doing, we can most definitely sandbox the rendering of the template. I'm not sure if `jinjava` sandboxes by default but I suspect this problem is also present in Echo, as well.

ethanfrogers:
I'll address the rest of the comments today.

ezimanyi:
The template injection vulnerability isn't an issue in echo, as (at least to my knowledge) it never accepts user-supplied templates.  The templates that it processes are always set a configuration time (and in general I believe our strategy is to view operators with access to the config files as trusted) and only the data to bind in is supplied by the user---here both data and template are user-supplied.

ethanfrogers:
Ah, fair enough.

ethanfrogers:
Closing because I haven't had time/priority to address comments or security concerns.

PR #362, opened 2019-04-09
fix(bake/manifests): capture only stdout in Helm bakery - 2nd attempt

ngaya-ll:
This is a fix for https://github.com/spinnaker/spinnaker/issues/3063.

Previously, #269 attempted to fix this issue by capturing stdout and stderr separately, however this change caused a regression and was reverted in #279.

This fix takes a similar approach to #269 but avoids the regression by keeping stdout and stderr completely separate and not attempting to merge the streams.  Additionally, the change is only enabled when baking Helm manifests, so other bake functionality will not be impacted.

One limitation of this change is that non-fatal warnings printed to stderr will be ignored if the `helm template` command exits with error code 0.  (Previously, non-fatal warnings would also not be reported, but would be merged into the output artifact data.  Consequently, the bake stage would appear to succeed but subsequent deploy stages would fail.)

A possible future enhancement would be to report non-fatal warnings to the end user and optionally fail the stage if anything is printed to stderr.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
@@ -64,7 +64,7 @@ class JobExecutorLocal implements JobExecutor {
           ByteArrayOutputStream stdErr
           if (jobRequest.combineStdOutAndErr) {
             stdOut = new ByteArrayOutputStream()
-            stdErr = stdOut
+            stdErr = null
             pumpStreamHandler = new PumpStreamHandler(stdOut)
           } else {
             stdOut = new ByteArrayOutputStream()
@@ -149,7 +149,7 @@ class JobExecutorLocal implements JobExecutor {
         }
 
         String outputContent = new String(stdOut.toByteArray())
-        String logsContent = stdOut.is(stdErr) ? outputContent : new String(stdErr.toByteArray())
+        String logsContent = (stdErr == null) ? outputContent : new String(stdErr.toByteArray())
 
         if (resultHandler.hasResult()) {
           log.info("State for $jobId changed with exit code $resultHandler.exitValue.")
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
@@ -59,8 +59,18 @@ class JobExecutorLocal implements JobExecutor {
       new Action0() {
         @Override
         public void call() {
-          ByteArrayOutputStream stdOutAndErr = new ByteArrayOutputStream()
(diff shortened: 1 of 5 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
```diff
@@ -59,8 +59,18 @@ class JobExecutorLocal implements JobExecutor {
...
         @Override
         public void call() {
-          ByteArrayOutputStream stdOutAndErr = new ByteArrayOutputStream()
-          PumpStreamHandler pumpStreamHandler = new PumpStreamHandler(stdOutAndErr)
+          PumpStreamHandler pumpStreamHandler
+          ByteArrayOutputStream stdOut
+          ByteArrayOutputStream stdErr
+          if (jobRequest.combineStdOutAndErr) {
+            stdOut = new ByteArrayOutputStream()
+            stdErr = stdOut
```
ezimanyi: I'm wondering if it would be clearer to just have `stdErr` be null in this case rather than have two references to the same stream.  The check `stdOut.is(stdErr)` below would then just be `stdErr == null` which also seems to convey the intention better.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
```diff
@@ -45,7 +45,7 @@ public Artifact bake(HelmBakeManifestRequest request) {
     BakeStatus bakeStatus;
 
     try {
-      JobRequest jobRequest = new JobRequest(recipe.getCommand(), new ArrayList<>(), UUID.randomUUID().toString());
+      JobRequest jobRequest = new JobRequest(recipe.getCommand(), new ArrayList<>(), UUID.randomUUID().toString(), /* combineStdOutAndErr*/ false);
```
ezimanyi: Minor, but we usually don't add comments like this; the function definition is documentation of what the parameter does.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy
ngaya-ll: Sure, that seems reasonable.

ngaya-ll:
The bug in the previous PR was that it attempted to copy the data from stdout and stderr during the code that first launches the job, after a hard-coded 500ms sleep.  See  ([here|https://github.com/spinnaker/rosco/pull/269/files#diff-be44ad8037e48cc6c69c2d7219b31f0cR106]).  As a result, any subsequent output after 500ms would not be copied to the combined `stdOutAndErr` stream used by the other bake tasks.

This PR avoids the issue by (1) not merging the streams and (2) performing all buffer copying in the `updateJob()` method.  It would be fix the bug in the previous PR, but I think merging the streams is usually not necessary or even desirable.  For baking Helm charts, at least, there is a clear separation between the Kube manifest output on stdout and logging/error messages on stderr.

PR #366, opened 2019-04-22
perf(redis): Stop using KEYS command in BakeStore

danopia:
Replace `jedis.keys` invocations with a kork `withKeyScan` invocation, like https://github.com/spinnaker/igor/pull/270 (instead of implementing a scan loop directly). Came up from Redis slowlog entries reaching dozens of ms. 

Note that I did not change out the other `jedisPool` usages in `RedisBackedBakeStore.groovy`. Replacing them introduced a bunch of these 'ambiguous call' warns:
Screen Shot 2019-04-22 at 9 56 18 AM

I was also unable to locate a `jedis.time()` signature for `getTimeInMilliseconds()`. The 'TIME' command is seemingly only exposed directly on BinaryClient.

For these reasons I only touched the 'keys' invocations in this initial commit.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.rosco.persistence
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate
 import com.netflix.spinnaker.rosco.api.Bake
 import com.netflix.spinnaker.rosco.api.BakeRequest
 import com.netflix.spinnaker.rosco.api.BakeStatus
@@ -37,11 +38,13 @@ class RedisBackedBakeStore implements BakeStore {
 
   private ObjectMapper mapper = new ObjectMapper()
   private JedisPool jedisPool
+  private RedisClientDelegate redisClientDelegate
 
   def scriptNameToSHAMap = [:]
 
-  public RedisBackedBakeStore(JedisPool jedisPool) {
+  public RedisBackedBakeStore(JedisPool jedisPool, RedisClientDelegate redisClientDelegate) {
     this.jedisPool = jedisPool;
+    this.redisClientDelegate = redisClientDelegate;
   }
 
   private void cacheAllScripts() {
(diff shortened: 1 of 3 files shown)


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -453,29 +456,17 @@ class RedisBackedBakeStore implements BakeStore {
...
   public String deleteBakeByKey(String bakeKey) {
-    def jedis = jedisPool.getResource()
     def keyList = [bakeKey, "allBakes"]
-
-    jedis.withCloseable {
-      Set<String> incompleteBakesKeys = jedis.keys(allIncompleteBakesKeyPattern)
-
-      keyList += incompleteBakesKeys
-    }
+    keyList += this.scanIncompleteBakesKeys()
```
cfieber: minor stylistic nit - you don't need the explicit `this` qualifier to call the scanIncompleteBakesKeys method, can just write this as
```groovy
keyList += scanIncompleteBakesKeys()
```

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -453,29 +456,17 @@ class RedisBackedBakeStore implements BakeStore {
...
-    def jedis = jedisPool.getResource()
     def updatedTimestampMilliseconds = timeInMilliseconds
     def keyList = [bakeKey, "allBakes"]
-
-    jedis.withCloseable {
-      Set<String> incompleteBakesKeys = jedis.keys(allIncompleteBakesKeyPattern)
-
-      keyList += incompleteBakesKeys
-    }
+    keyList += this.scanIncompleteBakesKeys()
```
cfieber: same comment as above

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
```diff
@@ -489,15 +480,9 @@ class RedisBackedBakeStore implements BakeStore {
...
-    def jedis = jedisPool.getResource()
     def updatedTimestampMilliseconds = timeInMilliseconds
     def keyList = [bakeId, "allBakes"]
-
-    jedis.withCloseable {
-      Set<String> incompleteBakesKeys = jedis.keys(allIncompleteBakesKeyPattern)
-
-      keyList += incompleteBakesKeys
-    }
+    keyList += this.scanIncompleteBakesKeys()
```
cfieber: same comment as above

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/RedisBackedBakeStore.groovy
danopia: Thanks! I'm not sure I fully registered that this is a Groovy file 🙈

PR #371, opened 2019-05-15
chore(logging): propagate and log orca executionId

asher:


rosco-core/src/test/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocalSpec.groovy
```diff
@@ -65,7 +65,7 @@ class JobExecutorLocalSpec extends Specification implements TestDefaults {
     when:
       def jobId = jobExecutorLocal.startJob(jobRequest)
       // Give the script time to run + 100 ms fudge factor
-      sleep(1100)
+      sleep(3000)
```
asher: This is a not great temporary workaround to this test being flakey on TravisCI with the prior sleep value.

PR #373, opened 2019-05-17
fix(artifacts): Fix NPE on empty artifact

ezimanyi:
Fixes spinnaker/spinnaker#4387

If an artifact is empty, we throw an NPE when downloading it rather than just pass an empty artifact along to the caller.

In general it's not overly useful to have an empty artifact, but in some cases (ex: a values.yaml file for a Helm bake) it's reasonable to have no content.


ezimanyi:
@spinnakerbot cherry-pick 1.14

PR #389, opened 2019-06-21
refactor(core): Convert package name logic to java

ezimanyi:
When trying to upgrade to Spring Boot 2 (and along with it groovy 2.5) I learned that Groovy changed the .pop() method from removing the last element to removing the first element.

This broke some of the package name logic, which used the .pop() method. Sidestep this entirely by converting it to java, as it's well-tested enough this can be done with reasonable confidence.


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/PackageNameConverter.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2015 Google, Inc.
```
maggieneterval:
```suggestion
 * Copyright 2019 Google, Inc.
```

ezimanyi: I left that date in that file as technically I only moved it `.groovy -> .java` though enough lines changed that github didn't catch it.  (The other files didn't have headers to start so I just put 2019.)

maggieneterval: Oh I see I see!

PR #394, opened 2019-06-26
chore(*): Apply codestyles

ezimanyi:
Enable codestyle enforcement on all subprojects and commit the results of spotlessApply.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,10 +53,6 @@ subprojects {
       exceptionFormat = 'full'
     }
   }
-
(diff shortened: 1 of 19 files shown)


ezimanyi:
Did this separately from the boot 2 upgrade to reduce the diff, though this is a relatively small diff because we don't apply codestyles to groovy and most of rosco is still groovy.

maggieneterval:
(Not a great emoji, was going for "style")

PR #407, opened 2019-08-13
chore(packer): upgrade packer binary to 1.3.2

jingsfdc:
Upgrades packer binary to 1.3.2 so we can take advantage of multi-region and HA bakes through the use of subnet_filter and vpc_filter.

https://github.com/spinnaker/spinnaker/issues/4114

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,9 +10,9 @@ RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
   dpkg -i ./rosco-web/build/distributions/*.deb && \
   mkdir /packer && \
   cd /packer && \
-  wget https://releases.hashicorp.com/packer/1.3.1/packer_1.3.1_linux_amd64.zip && \
+  wget https://releases.hashicorp.com/packer/1.3.2/packer_1.3.2_linux_amd64.zip && \
   apt-get install unzip -y && \
-  unzip packer_1.3.1_linux_amd64.zip && \
+  unzip packer_1.3.2_linux_amd64.zip && \
   rm -rf /workdir
 
 RUN wget https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get && \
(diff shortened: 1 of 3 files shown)


clanesf:
@ezimanyi - What are the chances of us getting this cherry-picked back into the next 1.14 release? We're not quite ready for 1.15 yet.

ezimanyi:
@clanesf : My only worry about cherry-picking this is that we have seen issues with upgrading packer in the past causing bugs due to API changes (ex: #310) and I don't want to cause a regression in a stable build.

That being said, this upgrade is just across a patch release of packer to it's probably reasonably safe to also do that upgrade in a patch release of Spinnaker.  What change is is that you need from Packer 1.3.2---is it the multi-region bake functionality in the description above?

clanesf:
@ezimanyi - yep, the multi-region bake is what we need.

matthack:
Sorry about being late to the party, but is is possible to have support for at least packer 1.4.1? We need packer's support for omitting volumes from the ebssurrogate packer provider. We use this to create a sparse image as recommended by aws so that our published ami's come up faster.

https://aws.amazon.com/blogs/apn/how-to-build-sparse-ebs-volumes-for-fun-and-easy-snapshotting/

We have had to run 1.4.1 in order to support this and have been running into an issue. Chiefly that certain bakes start on rosco but it looks like whatever listener is being used times out and marks the bake as failed, a complete image still is produced ironically enough. (https://github.com/spinnaker/spinnaker/issues/4498)

ezimanyi:
@matthack : I'd definitely be fine upgrading to 1.4.1 on master (and thus to be released in Spinnaker 1.16.0).  Actually given that 1.4.2 is out it probably makes sense to go straight to that version.  Is this something you'd mind making a PR for (which would of course be almost identical to this one)?

matthack:
@ezimanyi I'm happy to submit a PR, albeit a bit hesitant to. FWIW, we started having issues these issues with bakes after going to 1.4.1. Perhaps this will help flush them out.

PR #409, opened 2019-08-13
chore(packer): upgrade packer binary to 1.3.5

jingsfdc:
Turns out packer 1.3.2-1.3.4 have a lot of crash issues and builders filtering do not work.

spinnaker/spinnaker#4114


jingsfdc:
@ezimanyi pls and ty again

PR #410, opened 2019-08-14
chore(packer): upgrade packer binary to 1.4.2

matthack:
Upgrading to the latest current version of packer. We've ran 1.4.1 internally for a while but have run into issues with bakes, not sure if this is related to the upgrade.

https://github.com/spinnaker/spinnaker/issues/4498

spinnaker/spinnaker#4114


matthack:
Upgrade looks to pull in features from 1.4.1 that allow discarding of volumes with the ebsurrogate provider from the final image.

cc @ezimanyi

ezimanyi:
CC @jingsfdc @clanesf : Any concerns with bumping packer all the way to 1.4.2?  I believe you said 1.3.2-1.3.4 had issues, not sure if you have experience with 1.4.2.

jingsfdc:
@ezimanyi  Our OS team said it should be ok but they haven't tested. Can we have 1.3.5 in 1.14.x and have 1.4.2 as part of 1.15.x?

ezimanyi:
@jingsfdc : I'm a bit wary of bumping packer in a patch release, particularly if there have been issues with stability of packer recently.  I'll merge this bump to 1.4.2 into `master` (which means it will get released in 1.16.0).  How confident are you that we won't introduce regressions for Spinnaker users on 1.14.x and 1.15.x if we bump it to 1.13.5 there?  (I don't have a great sense of how risky this is myself.)

PR #413, opened 2019-08-16
feat(aws/ami_regions): add support for copies of amis

gardleopard:
With packer one can copy amis from one bake to different regions. This
way one can have one bake and use the same artifact in different
regions.

In the issue spinnaker/spinnaker#4641 this feature is requested.

This pull request keeps backwards compatability. It starts collecting
all amis that are produced in a bake and stores them in a list. It adds
a field `regions` to the api that will allow spinnaker/orca to pick up
that we have amis for different regions. In the bake endpoint in rosco
we start using region to filter through the various amis in artifacts if
it is present.

There needs to be done changes to orca as well in order to use this.


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -170,9 +174,10 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
   Bake scrapeCompletedBakeResults(String region, String bakeId, String logsContent) {
     String amiId
     String imageName
-
+    List<Artifact> artifacts = new ArrayList()
```
jervi: Should be `new ArrayList<>()`

gardleopard:
example from rosco api
```
{
	"id": "bf97957e-d6f3-460b-bb28-7490af707d95",
	"ami": " ami-0647884bbba9855be",
	"image_name": null,
	"regions": [
		"eu-north-1",
		"eu-west-1"
	],
	"artifact": {
		"type": "aws/image",
		"customKind": false,
		"name": "vim-all-20190816081314-trusty",
		"version": null,
		"location": "eu-west-1",
		"reference": " ami-0647884bbba9855be",
		"metadata": null,
		"artifactAccount": null,
		"provenance": null,
		"uuid": "bf97957e-d6f3-460b-bb28-7490af707d95"
	}
}
```

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -241,7 +241,16 @@ class BakeryController {
     def bake = bakeStore.retrieveBakeDetailsById(bakeId)
 
     if (bake) {
-      return bake
+      if (bake.artifacts?.size() > 0) {
+        bake.artifact = bake.artifacts.find{ it.location == region}
+        bake.regions = bake.artifacts.collect{it.location}
```
jervi: Can be replaced with `bake.artifacts*.location`

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
```diff
@@ -241,7 +241,16 @@ class BakeryController {
...
+        bake.ami = bake.artifact.reference
+        bake.artifacts = null
+
+        return bake
+      } else {
+        return bake
+      }
```
jervi: Drop the `else` and just always return bake

gardleopard: fixed

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/BakeryController.groovy
gardleopard: fixed

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
gardleopard: fixed

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -33,6 +34,9 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
 
   private static final String IMAGE_NAME_TOKEN = "amazon-(chroot|ebs): Creating the AMI:"
   private static final String UNENCRYPTED_IMAGE_NAME_TOKEN = "(==> |)amazon-(chroot|ebs): Creating unencrypted AMI"
+  private static final String AMI_EXTRACTOR = "[a-z]{2}-[a-z]{2,10}-[0-9]:\\sami-[a-z0-9]+"
```
asher: Can you add a comment explaining this?

gardleopard: Thanks, done :)

german-muzquiz:
Was orca ever updated to support these changes? I still can see the bake happening in unwanted regions, but maybe I just don't know how to configure the stage to prevent this.

PR #418, opened 2019-08-21
fix(bake/helm): use --set instead of --set-string

ethanfrogers:
using `--set` instead of `--set-string` ensures that values of different types get injected into into the chart properly. the most common use case for this is values that are used as `bool` in templates. Previously, these values would cause errors because they weren't properly converted.


ethanfrogers:
spinnaker/spinnaker#3950

PR #421, opened 2019-08-27
feature(kustomize): add integration for kustomize

ethanfrogers:
implements support for kustomize as a templating engine. one of the
interesting challenges with kustomize is that it depends on files
adjacent to a single kustomization file for it's templating. the
kustomization file refers to files surrounding it. these files are
collected and layered on top of eachother to produce a single, fully
rendered manifest. because of this we've had to implement a special
handling of kustomization files. essentially, before we attempt to
render a kustomization we parse the file and generate artifact requests
for any dependencies. we'll subsequently do this for any dependencies
that happen to be kustomization files themselves.


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
```diff
@@ -0,0 +1,102 @@
...
+    for (String name : names) {
+      try {
+        Artifact testArtifact = artifactFromBase(artifact, artifactPath.resolve(name).toString());
+        k = convert(testArtifact);
+        k.setKustomizationFilename(name);
+        break;
+      } catch (Exception e) {
+        log.error(
+            "kustomization file cannot be found for {}", artifactPath.resolve(name).toString());
+        continue;
```
ezimanyi: This `continue` isn't really necessary, as the last statement in the loop.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
```diff
@@ -0,0 +1,102 @@
...
+  private final RetrySupport retrySupport = new RetrySupport();
+  private static final List<String> KUSTOMIZATION_FILENAMES =
+      ImmutableList.of("kustomization.yaml", "kustomization.yml", "kustomization");
+
+  public KustomizationFileReader(ClouddriverService clouddriverService) {
+    this.clouddriverService = clouddriverService;
+  }
+
+  public Kustomization getKustomization(Artifact artifact, String possibleName) {
+    Path artifactPath = Paths.get(artifact.getReference());
```
ezimanyi: I'm a bit confused by the logic in this function...where is the `possibleName` coming from?  It looks like we're trying all of the `KUSTOMIZATION_FILENAMES` but using an input String as a hint for what to try first?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
```diff
@@ -0,0 +1,102 @@
...
+
+    Kustomization k = null;
+    for (String name : names) {
+      try {
+        Artifact testArtifact = artifactFromBase(artifact, artifactPath.resolve(name).toString());
+        k = convert(testArtifact);
+        k.setKustomizationFilename(name);
+        break;
+      } catch (Exception e) {
+        log.error(
```
ezimanyi: Is the idea here that the first existing file from the above list ends up being set as the file name?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
```diff
@@ -0,0 +1,102 @@
...
+  private Artifact artifactFromBase(Artifact artifact, String path) {
+    return Artifact.builder()
+        .reference(path)
+        .artifactAccount(artifact.getArtifactAccount())
+        .type(artifact.getType())
+        .build();
+  }
+
+  private Kustomization convert(Artifact artifact) throws IOException {
+    // TODO(ethanfrogers): figure out how to use safe constructor here.
```
ezimanyi: I'm not an expert, but does the current form prevent deserialization code injection?  I think that from a security perspective we'll need to solve this before merging (if this version hasn't already solved it).

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeBakeManifestService.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.Base64;
+import java.util.UUID;
+import org.springframework.stereotype.Component;
+
+@Component
+public class KustomizeBakeManifestService
+    implements BakeManifestService<KustomizeBakeManifestRequest> {
+  private final KustomizeTemplateUtils kustomizeTemplateUtils;
+  private final JobExecutor jobExecutor;
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
ezimanyi: Looks unused now

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -0,0 +1,198 @@
...
+  public BakeRecipe buildBakeRecipe(
+      BakeManifestEnvironment env, KustomizeBakeManifestRequest request) {
+    BakeRecipe result = new BakeRecipe();
+    result.setName(request.getOutputName());
+    Artifact artifact = request.getInputArtifact();
+    if (artifact == null) {
+      throw new IllegalArgumentException("Exactly one input artifact must be provided to bake.");
+    }
+    String kustomizationfilename = FilenameUtils.getName(artifact.getReference());
+    if (kustomizationfilename != null
```
ezimanyi: Does it make sense to continue here if the file name is null, or should we throw in that case too?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
ethanfrogers: Correct.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/mapping/ConfigMapGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class ConfigMapGenerator {
+
+  private String name;
+
+  private List<String> files = null;
```
ezimanyi: Generally wouldn't explicitly initialize `null` here

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/mapping/Kustomization.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private List<String> patchesStrategicMerge = new ArrayList<>();
+
+  private List<PatchesJson6902> patchesJson6902 = new ArrayList<>();
+
+  private String kustomizationFilename;
+
+  @JsonIgnore private Map<String, Object> additionalProperties = new HashMap<>();
+
+  public HashSet<String> getFilesToEvaluate() {
```
ezimanyi: Suggest using `Set` for public method

rosco-manifests/src/test/groovy/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReaderSpec.groovy
```diff
@@ -0,0 +1,61 @@
+package com.netflix.spinnaker.rosco.manifests.kustomize
```
ezimanyi: Missing header

rosco-manifests/src/test/groovy/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtilsSpec.groovy
```diff
@@ -0,0 +1,195 @@
...
+                .name("examples/multibases/kustomization.yaml")
+                .reference(referenceBase + "examples/multibases/kustomization.yaml")
+                .artifactAccount("github")
+                .build()
+
+        when:
+        def filesToFetch = kustomizationTemplateUtils.getFilesFromArtifact(baseArtifact)
+
+        then:
+        7 * kustomizationFileReader.getKustomization(_ as Artifact, _ as String) >> { Artifact a, String s ->
```
ezimanyi: I'd suggest dropping the precise number of times this is called from the test, as this often leads to overspecified tests.  For mutating operations it's often important that we know exactly how many times they are called---but for more of a 'getter' like this it's probably better to just use a stub (that returns test data but doesn't care about number of calls) instead of a mock that enforces the exact number of calls.

rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/controllers/V2BakeryController.java
```diff
@@ -13,10 +17,28 @@
 @RestController
 @Slf4j
 public class V2BakeryController {
-  @Autowired HelmBakeManifestService helmBakeManifestService;
+  List<BakeManifestService> bakeManifestServices;
```
ezimanyi: These member variables can/should be `private final`

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -0,0 +1,198 @@
...
+  private Artifact artifactFromBase(Artifact artifact, String reference, String name) {
+    return Artifact.builder()
+        .reference(reference)
+        .name(name)
+        .type(artifact.getType())
+        .artifactAccount(artifact.getArtifactAccount())
+        .build();
+  }
+
+  private boolean isFolder(String evaluate) {
```
ezimanyi: I'm not sure I understand this logic...is this trying to determine if a path as a string represents a folder?  It feels like it should have some tests on it to make it clear it works

ezimanyi:
OK, gave it another pass, mostly looking good.  I think from my perspective the main blocker is just ensuring that we're safely deserializing YAML so we don't create any security issues.

The logic to download files looks a lot cleaner now.  Is there a design document or some artifact describing how this is designed?  It thing seeing it at a higher level might make it easier for me to understand the code.  I've left a few questions still on parts I don't fully understand.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
ethanfrogers: Correct. A Kustomization file can have 3 names, `.yml`, `.yaml` and one w/o an extension. Initially, the file name comes from the artifact that you supply as an input. We're making the assumption that if you use, say `kustomization.yml`, you're likely to use `kustomization.yml` for your subsequent kustomizations. Since we have no way of knowing that the actual file name is before we try to fetch it we try to be smart and make a best guess before falling back to the other options.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
ethanfrogers: I'm honestly not sure. I know `SafeConstructor` only lets you deserialize to Java primitives so using it here will cause an error.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
ethanfrogers: Yeah, it's probably safer to guard against a non well formed artifact than it is to assume we'll always be supplied a well formed artifact.

ethanfrogers:
@ezimanyi I can certainly document this in a more permanent location but will describe the design here.

Generally, a "kustomization"  is a folder, in a repo, and is made up of 2 things: 
1) A `kustomization` file which describes how the resulting manifest(s) should be rendered. This includes things like files to include (`resources`), patch files to modify those files (various `patch` keys), `configMapGenerators` which take various files and turn them into a ConfigMap, etc. 
2) All of the files that the `kustomization` file depends on.

It's also worth pointing out that a `kustomization` file can depend on other `kustomization`s in sibling or child directories e.x `../sibling` or `./child`.

Kustomize is typically used in CI and meant to be run against a full repo where all of the file dependencies are checked out alongside the `kustomization` file that defines them. However, since Spinnaker doesn't have an artifact that can represent this (a `git/repo` artifact, for example) we needed to come up with a way to ensure all of the file dependencies are present before we can run `kustomize build`.

Our solution is to treat each `kustomization` file as a BOM, essentially. For every `kustomization` we fine, we collect the files we know _aren't) references to other `kustomizations` and then attempt to fetch subsequent `kustomizations` and do the same thing. Once we have a full list of dependencies, we can reconstruct the file tree, write each file to disk and then run `kustomize build`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReader.java
ethanfrogers: For reference - Yes, using a `new Constructor(Kustomization.class)` will prevent code injection according to [this](https://lgtm.com/blog/swagger_snakeyaml_CVE-2017-1000207_CVE-2017-1000208).

ethanfrogers:
@ezimanyi I'd love to see this land in 1.16 so LMK if there's anything else we need to do.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -0,0 +1,199 @@
...
+  protected void downloadArtifactToTmpFileStructure(
+      BakeManifestEnvironment env, Artifact artifact, String referenceBaseURL) throws IOException {
+    if (artifact.getReference() == null) {
+      throw new InvalidRequestException("Input artifact has an empty 'reference' field.");
+    }
+    Path artifactFileName = Paths.get(extractArtifactName(artifact, referenceBaseURL));
+    Path artifactFilePath = env.getStagingPath().resolve(artifactFileName);
+    // ensure file write doesn't break out of the staging directory ex. ../etc
+    Path artifactParentDirectory = artifactFilePath.getParent();
+    if (!artifactParentDirectory.startsWith(env.getStagingPath())) {
```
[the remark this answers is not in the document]
ezimanyi: Actually I'm fairly certain this doesn't work as is; the following will return `true`:
```
    Path test = Paths.get("/tmp/mydir");
    Path other = Paths.get("/tmp/mydir/../../etc/passwd");
    boolean matches = other.startsWith(test);
```

ethanfrogers:
@ezimanyi agreed that this isn't least confusing solution out there. We opted for this solution for a couple of reasons:

1) It's as close to the normal `kustomize` experience as we can get with what we have today. Your suggestion about expecting a `.tar` is reasonable, however, as a user of `kustomize` I wouldn't see much value in using it within Spinnaker if I had to go the extra step and package up my code before I could. That would happen in CI and, if I already have Kustomize there I might as well do it at that stage.
2) Adding a `git/repo` or folder structure artifact would have increased the scope of this change dramatically. I agree that we _should_ have something that represents that in the future but this seemed like a reasonable solution until we can agree on what that would look like. 

I'm happy to call this feature "beta" for the 1.16 release. I've already put it behind a feature flag in Deck. Would this solution be reasonable if the logic were hardened/tested more? If so, is this something you'd feel comfortable merging or are we calling this a "no-go"?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -0,0 +1,199 @@
...
+    // example: https://api.github.com/repos/org/repo/contents/kustomize.yml == kustomize.yml
+    return artifact.getReference().replace(artifact.getName(), "");
+  }
+
+  /**
+   * getFilesFromArtifact will use a single input artifact to determine the dependency tree of files
+   * mentoined it's (and subsequent) kustomization file.
+   */
+  private HashSet<String> getFilesFromArtifact(Artifact artifact) throws IOException {
+    // referenceBaseURL is the base URL without the artifacts name. works for github and bitbucket
```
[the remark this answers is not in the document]
ethanfrogers: What if we limited the input artifact types to Github and Bitbucket artifacts? I agree that it isn't ideal but maybe we can use this as an experiment to determine what artifact types people use this with most often. If people use it with more than Github or Bitbucket we can revisit.

ezimanyi:
I think the solution is reasonable; while it might have been good to think a bit more broadly, I don't think this is a bad solution, I just wonder a bit about the tradeoffs between the options noted above.  That was more of a general comment about wishing we'd have a more open design document as I read the complexity of the implementation, not something I think we need to address/fix here.

From a more tactical perspective about this in 1.16, my main concern is just that I can't really convince myself this is correct---I've read it through a few times and am still having trouble understanding exactly what it's doing (or finding enough tests to convince me).  At a minimum I'd want to see the few bugs noted in the latest review fixed before merging this...but in particular the one about dropping a slash from all URLs (`https:/my-url/`) makes me wonder how much this has been tested.  Until that stuff is fixed, I'd say this isn't even at something I'd call "beta" yet, but with that and some more tests to validate the fixes to those bugs then this could be mergeable.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -0,0 +1,199 @@
...
+    return Artifact.builder()
+        .reference(reference)
+        .name(name)
+        .type(artifact.getType())
+        .artifactAccount(artifact.getArtifactAccount())
+        .build();
+  }
+
+  private boolean isFolder(String evaluate) {
+    if (evaluate.contains(".")) {
```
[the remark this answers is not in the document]
ethanfrogers: Moved to just checking for an extension on the file via `FilenameUtils.getExtension`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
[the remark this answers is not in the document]
ethanfrogers: I see what we need to do here - we need a call to `normalize` before we to the assertion.

PR #429, opened 2019-09-09
fix(bake/helm): Remove tests directory when baking Helm chart

Aaquiff:
When baking Helm charts using the command `helm template`, ignore the templates in the tests in the directory `<CHART>/templates/tests`.

Resolves [issue](https://github.com/spinnaker/spinnaker/issues/4831)

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
@@ -37,7 +37,9 @@ public class HelmBakeManifestService extends BakeManifestService<HelmBakeManifes
       return Artifact.builder()
           .type("embedded/base64")
           .name(bakeManifestRequest.getOutputArtifactName())
-          .reference(Base64.getEncoder().encodeToString(bakeResult))
+          .reference(
+              Base64.getEncoder()
+                  .encodeToString(helmTemplateUtils.removeTestsDirectoryTemplates(bakeResult)))
           .build();
     }
   }
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmTemplateUtils.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmTemplateUtils.java
@@ -14,12 +14,17 @@ import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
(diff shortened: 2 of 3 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
```diff
@@ -63,6 +71,25 @@ public BakeManifestService(JobExecutor jobExecutor) {
     if (bakeStatus.getResult() != BakeStatus.Result.SUCCESS) {
       throw new IllegalStateException("Bake failed: " + bakeStatus.getLogsContent());
     }
-    return bakeStatus.getOutputContent().getBytes();
+    return removeTestsDirectoryTemplates(bakeStatus.getOutputContent()).getBytes();
+  }
+
+  public String removeTestsDirectoryTemplates(String input) {
```
ethanfrogers: I think this probably makes more sense in `HelmBakeManifestService` since test manifests are only a concept within Helm.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
```diff
@@ -63,6 +71,25 @@ public BakeManifestService(JobExecutor jobExecutor) {
...
+            .filter(
+                manifest ->
+                    !manifest.trim().isEmpty()
+                        && !Pattern.compile(REGEX_TESTS_MANIFESTS).matcher(manifest).find())
+            .collect(Collectors.toList());
+
+    final StringBuilder stringBuilder = new StringBuilder();
+    outputManifests.stream()
+        .forEach(manifest -> stringBuilder.append(MANIFEST_SEPARATOR).append(manifest));
+    return stringBuilder.toString();
```
ethanfrogers: I think this can be simplified using streams and a joining collector. I haven't tried it but it might be a bit easier to read.
```
String manifestBody = outputManifests.stream().collect(Collectors.joining(MANIFEST_SEPARATOR + "\n"))
return manifestBody.getBytes()
```

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
Aaquiff: Moved to `HelmBakeManifestService `

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestService.java
Aaquiff: Fixed

PR #435, opened 2019-09-25
feat(packer): set custom repository per base image

sergiopena:
* feat(packer): set custom repository per base image

Right now there is the posibility to add one repository per package type (deb/rpm)
this collides when trying to setup multiple flavours/versions repositories for the
same package type.

e.g. When baking with Ubuntu you can only setup one repository index (xenial, bionic)
this forces you to put in the same repository index all packages no matter which
version are they intented to be.

With this patch, you can add a customRepository property per BaseImage with overwrites the
debianRepository/yumRepository/chocolateyRepository global config entries, this provides
more granularity on the repository configuration.

https://github.com/spinnaker/spinnaker/issues/4894


robzienert:
Would like a test around this functionality.

sergiopena:
Hi @robzienert 👋 

Thanks a lot for reviewing this PR, I've added unit test for debian images with the custom repository set.

PR #439, opened 2019-10-07
fix(bakeManifest): revert to set-string

ethanfrogers:
reverts a change introduced in #418 where we went from using
`--set-string` to `--set`. This change caused backwards incompaitble
changes because the parser for `--set` in helm coerces integers into
floats when it injects them into templates. `--set-string` was
introduced to fix this in helm 2.


ethanfrogers:
spinnaker/spinnaker#3950

ethanfrogers:
Once merged we can backport to 1.16 and I'll follow up with a PR to preserve the `--set` functionality in 1.16 and forward.

ethanfrogers:
@spinnakerbot cherry-pick 1.16

PR #441, opened 2019-10-08
fix(bakeManifest): helm --set option

ethanfrogers:
adds an option to toggle between `--set` and `--set-string`. we recently
had to revert a commit which changed `--set-string` to `--set` and
introduced a breaking change. this keeps the default behavior of
`--set-string` but allows users to choose which option they want to use.

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestRequest.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestRequest.java
@@ -11,4 +11,5 @@ import lombok.EqualsAndHashCode;
 public class HelmBakeManifestRequest extends BakeManifestRequest {
   String namespace;
   List<Artifact> inputArtifacts;
+  boolean rawOverrides;
 }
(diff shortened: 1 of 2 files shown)


maggieneterval:
@ethanfrogers are there corresponding Orca/Deck PRs for configuring thist option? Also curious if you think there will ever potentially be options besides `--set` and `--set-string` and we should use an enum rather than a boolean here?

ezimanyi:
It looks like this is only going to be a top-level option for the entire manifest bake---will this work for people who want to set both integers and booleans as overrides?  I'm wondering if we should think about a solution that will work for that case, as it will be more confusing to implement a solution to that case if we already have a top-level toggle between set and set string for the entire stage.

ethanfrogers:
@ezimanyi I share the same concern. However, I think we have a couple of options. 

1. If we want to add two separate sections we could toggle off of this property and fill out the appropriate section (overrides or rawOverrides) in the UI. 
2. We could move to a more complex data structure to represent each individual value (key, value, raw|string), which I think would be possible, given the current data structure (`Map<String, Object>).

Either way, I think we can _safely_ transition to either one, using this value to determine the type of override, at least in the UI. 

Either way, we'll have to support backwards compatibility

ethanfrogers:
@spinnakerbot 🍒 ⛏ 1.16

ethanfrogers:
@spinnakerbot cherry-pick 1.16

PR #442, opened 2019-10-10
fix(helm): Fix baking of helm artifacts

ezimanyi:
Fixes spinnaker/spinnaker#4874.

We currently throw an exception if a helm artifact doesn't have a reference. We're only using that reference to generate a file name to use when downloading the artifact; helm artifacts don't actually need a reference.

Fix this by just falling back to the name-version if there is no reference defined.

(I'd like to refactor this code a bit, and just use a UUID as the file name to avoid all this, but this felt like a smaller change and I'd like to cherry pick the fix.)


ezimanyi:
@spinnakerbot :cherries: :pick: 1.16

ezimanyi:
Manual :cherries: :pick: to 1.16 in #443 .

PR #446, opened 2019-10-15
fix(kustomize): fetch new artifacts using the same version

kevinawoo:
if it's not set, it'll fetch from `master`, which is unintended.


ethanfrogers:
Hopefully we land support for `git/repo` in 1.17 and won't need this :D

PR #449, opened 2019-10-17
feat(kustomize): support git/repo artifact in kustomize bake manifest

KathrynLewis:
This adds support for `git/repo` as an artifact type for a kustomize bake. This is a work in progress involving changes to other services as well, so we're leaving in the old logic (baking from a github/file artifact) until the rest of the PRs are merged. 

@ezimanyi PTAL?

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
@@ -29,22 +29,20 @@ public final class ArtifactDownloaderImpl implements ArtifactDownloader {
         retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
     if (response.getBody() == null) {
       throw new IOException("Failure to fetch artifact: empty response");
-    } else {
-      try (InputStream inputStream = response.getBody().in()) {
-        return inputStream;
+    }
+    return response.getBody().in();
+  }
+
+  public void downloadArtifactToFile(Artifact artifact, Path targetFile) throws IOException {
+    try (OutputStream outputStream = Files.newOutputStream(targetFile)) {
+      try (InputStream inputStream = downloadArtifact(artifact)) {
+        IOUtils.copy(inputStream, outputStream);
       } catch (IOException e) {
         throw new IOException(
             String.format(
                 "Failed to read input stream of downloaded artifact: %s. Error: %s",
                 artifact, e.getMessage()));
       }
-    }
-  }
-
-  public void downloadArtifactToFile(Artifact artifact, Path targetFile) throws IOException {
-    try (OutputStream outputStream = Files.newOutputStream(targetFile)) {
-      InputStream inputStream = downloadArtifact(artifact);
-      IOUtils.copy(inputStream, outputStream);
     } catch (RetrofitError e) {
       throw new IOException(
           String.format("Failed to download artifact: %s. Error: %s", artifact, e.getMessage()));
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
@@ -27,28 +27,24 @@ public final class ArtifactDownloaderImpl implements ArtifactDownloader {
   public InputStream downloadArtifact(Artifact artifact) throws IOException {
     Response response =
         retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
-    if (response.getBody() != null) {
-      return response.getBody().in();
+    if (response.getBody() == null) {
+      throw new IOException("Failure to fetch artifact: empty response");
     } else {
-      log.error("Failure to fetch artifact: empty response");
-      return null;
+      try (InputStream inputStream = response.getBody().in()) {
+        return inputStream;
+      } catch (IOException e) {
+        throw new IOException(
+            String.format(
+                "Failed to read input stream of downloaded artifact: %s. Error: %s",
+                artifact, e.getMessage()));
+      }
     }
   }
 
   public void downloadArtifactToFile(Artifact artifact, Path targetFile) throws IOException {
     try (OutputStream outputStream = Files.newOutputStream(targetFile)) {
-      Response response =
-          retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
-      if (response.getBody() != null) {
-        try (InputStream inputStream = response.getBody().in()) {
-          IOUtils.copy(inputStream, outputStream);
-        } catch (IOException e) {
-          throw new IOException(
-              String.format(
-                  "Failed to read input stream of downloaded artifact: %s. Error: %s",
-                  artifact, e.getMessage()));
-        }
-      }
+      InputStream inputStream = downloadArtifact(artifact);
+      IOUtils.copy(inputStream, outputStream);
     } catch (RetrofitError e) {
       throw new IOException(
           String.format("Failed to download artifact: %s. Error: %s", artifact, e.getMessage()));
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
@@ -23,7 +23,18 @@ public final class ArtifactDownloaderImpl implements ArtifactDownloader {
     this.clouddriverService = clouddriverService;
   }
 
-  public void downloadArtifact(Artifact artifact, Path targetFile) throws IOException {
+  public InputStream downloadArtifact(Artifact artifact) throws IOException {
+    Response response =
+        retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
+    if (response.getBody() != null) {
+      return response.getBody().in();
+    } else {
+      log.error("Failure to fetch artifact: empty response");
+      return null;
+    }
+  }
+
+  public void downloadArtifactToFile(Artifact artifact, Path targetFile) throws IOException {
     try (OutputStream outputStream = Files.newOutputStream(targetFile)) {
       Response response =
           retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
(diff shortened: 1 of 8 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -79,11 +99,71 @@ public BakeRecipe buildBakeRecipe(
...
+
+    Path tarPath = env.resolvePath("repo.tar.gz");
+    try {
+      artifactDownloader.downloadArtifact(artifact, tarPath);
+    } catch (IOException e) {
+      throw new IOException("Failed to download git/repo artifact: " + e.getMessage(), e);
+    }
 
+    try {
+      extractArtifact(tarPath, env.resolvePath(""));
```
ezimanyi: Currently, we're doing the following:
* Requesting the artifact from clouddriver, then writing that `InputStream` to a file in `artifactDownloader.downloadArtifact()`
* Opening the file as an `InputStream` and decompressing the archive

This would be simpler (and likely more efficient) if we directly consumed the `InputStream` we got back from clouddriver and decrypted from that, rather than needing the intermediate write to disk.

Perhaps we could rename `ArtifactDownloader.downloadArtifact` to `ArtifactDownloader.downloadArtifactToFile` and then have `ArtifactDownloader.downloadArtifact` just return an `InputStream`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -79,11 +99,71 @@ public BakeRecipe buildBakeRecipe(
...
+    command.add(env.resolvePath(kustomizeFilePath).getParent().toString());
+
+    BakeRecipe result = new BakeRecipe();
+    result.setCommand(command);
     return result;
   }
 
+  // This being here is temporary until we find a better way to abstract it
+  public static void extractArtifact(Path inputTar, Path outputPath) throws IOException {
+    TarArchiveInputStream tarArchiveInputStream =
```
ezimanyi: This stream won't get closed if there's an exception in the logic below; this can be fixed by using a try-with-resources construct:
```
try (TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(
            new GzipCompressorInputStream(new BufferedInputStream(Files.newInputStream(inputTar)))) {
//...
}
```

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -79,11 +99,71 @@ public BakeRecipe buildBakeRecipe(
...
+        new TarArchiveInputStream(
+            new GzipCompressorInputStream(new BufferedInputStream(Files.newInputStream(inputTar))));
+
+    ArchiveEntry archiveEntry = null;
+    while ((archiveEntry = tarArchiveInputStream.getNextEntry()) != null) {
+      Path archiveEntryOutput = outputPath.resolve(archiveEntry.getName());
```
ezimanyi: How does this code behave if the archive entry's name references the parent directory?  Could this break out of the temporary directory we're trying to extract to?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
KathrynLewis: Good catch--changed! But I'm not sure how we want to handle an empty response body in the downloadArtifact() method. Right now, I just have it logging an error and returning null.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
KathrynLewis: I tested this out and you're right that if I force an entry to reference a parent, it does break out. Is this something that we want to support and handle in some way? Or should it be considered an error if an entry at the repo level tries to reference a parent? Because in theory, we wouldn't even have access to those files, correct?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
ezimanyi: Good question!  I'd suggest throwing an `IOException` (with the text you're currently logging) if the response body is empty.  I realize that's a change from what we currently do, where we just silently don't write out the artifact to a file if there's no input, but it seems like an improvement as the error would be surfaced earlier and more clearly.  (As opposed to currently where it would just manifest as a missing file somewhere downstream.)

On the topic of errors, @maggieneterval just improved the error handling in this file in #453 so that errors actually propagate back to the user (per the screen shots on that PR).  So you'll probably have a minor merge conflict when you merge in her changes here.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
```diff
@@ -23,7 +23,18 @@ public ArtifactDownloaderImpl(ClouddriverService clouddriverService) {
...
+        retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
+    if (response.getBody() != null) {
+      return response.getBody().in();
+    } else {
+      log.error("Failure to fetch artifact: empty response");
+      return null;
+    }
+  }
+
+  public void downloadArtifactToFile(Artifact artifact, Path targetFile) throws IOException {
```
ezimanyi: To reduce code duplication, I'd suggest having `downloadArtifactToFile` call the new `downloadArtifact` function above instead of directly calling out to `clouddriverService`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
ezimanyi: I think we'd want to immediately throw an exception and abort if the entry tries to reference something outside of the repo directory.  This would avoid security vulnerabilities where someone could send a specially-crafted repo artifact that unpacks on top of system files.  (You're right that in theory important files shouldn't be accessible by the `rosco` process unless it's running as `root` which it shouldn't, but I still think it's better to fail fast before depending on these permissions being set correctly.)

A good example where we do this check elsewhere and fail fast is [here](https://github.com/spinnaker/rosco/blob/14b794b7b09ab2516bbb7169457c9dd2a8dda4bc/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestEnvironment.java#L53).  (That example is not in a re-usable place, but you could do something similar in the function you wrote.)

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
```diff
@@ -79,11 +99,72 @@ public BakeRecipe buildBakeRecipe(
...
+        if (archiveEntry.isDirectory()) {
+          if (!Files.exists(archiveEntryOutput)) {
+            Files.createDirectory(archiveEntryOutput);
+          }
+        } else {
+          Files.copy(tarArchiveInputStream, archiveEntryOutput);
+        }
+      }
+    } catch (Exception e) {
+      System.out.println(e);
```
ezimanyi: I'd suggest removing this `catch` block entirely and letting any exception propagate out of this method and bubble up---if we catch the exception here and print it to the logs, the end-user will lose any visibility into it.

mrampton:
Linking to the PR in clouddriver which added git/repo: https://github.com/spinnaker/clouddriver/pull/4049

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
KathrynLewis: Didn't meant to leave that in there. Thanks for catching that.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
KathrynLewis: Yep, that makes a lot of sense. 👍

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
KathrynLewis: Sounds good--updated.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
```diff
@@ -24,20 +24,27 @@ public ArtifactDownloaderImpl(ClouddriverService clouddriverService) {
...
-                  "Failed to read input stream of downloaded artifact: %s. Error: %s",
-                  artifact, e.getMessage()));
-        }
+  public InputStream downloadArtifact(Artifact artifact) throws IOException {
+    Response response =
+        retrySupport.retry(() -> clouddriverService.fetchArtifact(artifact), 5, 1000, true);
+    if (response.getBody() == null) {
+      throw new IOException("Failure to fetch artifact: empty response");
+    } else {
+      try (InputStream inputStream = response.getBody().in()) {
```
ezimanyi: I don't think we'll want a `try-with-resources` in this function, as it's directly returning the `InputStream` which should remain open for the caller (whereas `try-with-resources` will close it).

I think what this function should do is just `return response.getBody().in()` after the null check.  Then callers (including the below `downloadArtifactToFile`) should call this function as:
```
try (InputStream inputStream = downloadArtifact(artifact)) {
// ...use inputStream
}
```
That means we'll have nested `try-with-resources` blocks in `downloadArtifactToFile` but that's fine and the usual way of solving this.

KathrynLewis: Ah, that makes sense--updated. I have the catches nested as well, to preserve the logic of the error messages Maggie recently added.

PR #453, opened 2019-10-22
fix(artifact): improve error messaging for failed artifact downl…

maggieneterval:
Related to: https://github.com/spinnaker/spinnaker/issues/4752

When running a Bake (Manifest) stage with the Helm or Kustomize template renderers, Orca calls out to Rosco, which calls out to Clouddriver to download the input artifact. If the download fails, a non-specific "500" message is surfaced in Deck. This change adds two explicit catch blocks to the `ArtifactDownloaderImpl` so that the full artifact and error message are surfaced in Deck. I am still not sure why Retrofit and/or Orca's handling of Retrofit errors only surfaces "500" by default, since Clouddriver does throw a more explicit error when it fails to download an artifact -- it seems like when the error is passed between multiple microservices to get back to Orca, the message on the body of the Retrofit response is lost.

Before:
Xq78dvzCrRj

After:
GjwCF9fUGn5


maggieneterval:
LOVE the :amphora:

PR #470, opened 2019-12-02
fix(gce): Remove recently unpublished 'trusty' base image, add 'bionic'

ttomsu:
It appears the `ubuntu-1404-lts` image family was unpublished, causing our builds to break.  This will need to be cherry picked into 1.17, too.

### DIFF
--- a/rosco-web/config/rosco.yml
+++ b/rosco-web/config/rosco.yml
@@ -398,14 +398,6 @@ google:
     #   virtualizationSettings:
     #     sourceImage: ubuntu-1404-trusty-v20170517
     - baseImage:
-        id: trusty
-        shortDescription: v14.04
-        detailedDescription: Ubuntu Trusty Tahr v14.04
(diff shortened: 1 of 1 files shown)


rosco-web/config/rosco.yml
```diff
@@ -413,6 +405,14 @@ google:
         isImageFamily: true
       virtualizationSettings:
         sourceImageFamily: ubuntu-1604-lts
+    - baseImage:
+        id: bionic
+        shortDescription: v18.04
+        detailedDescription: Ubuntu Bionic Beaver v14.04
```
duftler:
```suggestion
        detailedDescription: Ubuntu Bionic Beaver v18.04
```

ttomsu:
Chatted with team on not breaking currently supported top level releases. Ended up with a slightly different approach for past releases: https://github.com/spinnaker/rosco/pull/471

PR #473, opened 2019-12-02
fix(gce): Hardcode Trusty GCE image in halconfig/images.yml

ttomsu:
Need to make the same change as https://github.com/spinnaker/rosco/pull/471 except in `halyard/` in order for releases to pick them up.


ttomsu:
Merging this since it is blocking build and validate of 1.17

PR #474, opened 2019-12-02
fix(gce): Remove Trusty GCE image in halconfig/images.yml, add Bionic

ttomsu:
Need to apply https://github.com/spinnaker/rosco/pull/470 to `halyard/images.yml` to have it work with releases.


ttomsu:
Merging this since it's blocking the build and validate of the `master` branch

PR #487, opened 2019-12-13
fix(release): switch back to Alpine, but don't use the openjdk container

plumpy:


plumpy:
See spinnaker/spinnaker#5204

PR #537, opened 2020-02-20
feat(helm): Implement Support for Selecting Between Helm2 and Helm3 B…

evansb:
Part of spinnaker/spinnaker#5474

This PR does the following

- Add HELM3 Template Renderer type
- Add indirection in `HelmTemplateUtils` to select Helm 3 binary if HELM3 renderer is chosen
- Allow both Helm binary path to be customised using e.g `helm.v2ExecutablePath`

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
@@ -22,6 +22,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties("helm")
 @Data
 public class RoscoHelmConfigurationProperties {
-  String v3ExecutablePath = "helm3";
-  String v2ExecutablePath = "helm";
+  private String v3ExecutablePath = "helm3";
+  private String v2ExecutablePath = "helm";
 }
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2020 Grab Holdings, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.rosco.manifests.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("helm")
+@Data
+public class RoscoHelmConfigurationProperties {
+  String v3ExecutablePath = "helm3";
+  String v2ExecutablePath = "helm";
+}
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
@@ -12,7 +12,8 @@ import org.springframework.stereotype.Component;
 @Component
 public class HelmBakeManifestService extends BakeManifestService<HelmBakeManifestRequest> {
   private final HelmTemplateUtils helmTemplateUtils;
-  private static final String HELM_TYPE = "HELM2";
+  private static final String HELM2_TYPE = "HELM2";
+  private static final String HELM3_TYPE = "HELM3";
 
   public HelmBakeManifestService(HelmTemplateUtils helmTemplateUtils, JobExecutor jobExecutor) {
     super(jobExecutor);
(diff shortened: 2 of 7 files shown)


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2020 Netflix, Inc.
```
ethanfrogers: s/Netflix/Grab

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmTemplateUtils.java
```diff
@@ -96,4 +103,12 @@ private Path downloadArtifactToTmpFile(BakeManifestEnvironment env, Artifact art
     artifactDownloader.downloadArtifactToFile(artifact, targetPath);
     return targetPath;
   }
+
+  private String getHelmExecutableForRequest(HelmBakeManifestRequest request) {
+    if (BakeManifestRequest.TemplateRenderer.HELM2.equals(request.getTemplateRenderer())) {
+      return helmConfigurationProperties.getV2ExecutablePath();
+    } else {
+      return helmConfigurationProperties.getV3ExecutablePath();
+    }
```
ethanfrogers: Small nit, no need for the `else` here since you're returning early in the `if`.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.netflix.spinnaker.rosco.manifests.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties("helm")
+@Data
+public class RoscoHelmConfigurationProperties {
```
ezimanyi: Do we expect users to need to override these executable paths?  We didn't provide the option when we only supported helm 2 baking; does adding support for helm 3 change this?  I ask because anything we add to configuration properties becomes a public contract that we need to indefinitely support so want to avoid adding things "just in case" when defaults will work.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
```diff
@@ -12,7 +12,8 @@
 @Component
 public class HelmBakeManifestService extends BakeManifestService<HelmBakeManifestRequest> {
   private final HelmTemplateUtils helmTemplateUtils;
-  private static final String HELM_TYPE = "HELM2";
+  private static final String HELM2_TYPE = "HELM2";
```
ezimanyi: Instead of adding additional strings to keep track of this, it might be cleaner to do something like:
```
private static ImmutableSet<TemplateRenderer> supportedTemplates = ImutableSet.of(TemplateRenderer.HELM2, TemplateRenderer.HELM3)
```
then `handles` could be `supportedTemplates.contains(TemplateRenderer.fromString(type))`.

That just reduces the number of places we have strings `HELM*` floating around given that we already have an enum to represent it.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmTemplateUtils.java
ezimanyi: Or maybe a `switch` here?  It's fine now, but this will start getting way more complex if we ever add any other condition.

evansb:
@ethanfrogers @ezimanyi fyi: added some indirection to handle minor differences between two versions, I missed this before.

```
Helm 3 -> helm template <name> <chart> <flags>
Helm 2 -> helm template <chart> --name <name>
```

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
evansb: Yeah, I was quite hesitant on doing this initially, the reason why this might be useful is to support Rosco deployments that does not use the official image. `helm3` is not the official binary name, it is renamed from `helm` so that we can install them side-by-side as they are not 100% backwards compatible.
ref: https://github.com/spinnaker/rosco/commit/5da1ba05b20593a24803e55fcc44fff18ae76136

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
evansb: Updated to use `ImmutableSet<String>` from the enum values, `TemplateRenderer.fromString(type)` is slightly more involved since it is not exception free.

evansb:
I think I can refactor the command construction in `buildBakeRecipe` to use something like `HelmCommandBuilder` in a new PR to make it more maintainable.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/config/RoscoHelmConfigurationProperties.java
ethanfrogers: I've run into quite a few situations where, even if users are running the OSS images, they need to override a particular executable to fix a bug more quickly than we can cut a release. In particular, this happens a lot with Packer. I'm a +1 for making this easier on users.

ezimanyi: Sounds good, fine leaving this in given that there are use cases 👍

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helm/HelmBakeManifestService.java
ezimanyi: Looks good, thanks!  Ideally we'd maybe even make an exception-free version `fromString` so we could still use the enum, but this is definitely better as it keeps it more encapsulated.

ethanfrogers:
@evansb let us know when you're ready to merge this!

evansb:
@ethanfrogers I think this should be ready to merge. I will test in our Spinnaker instances with `master-latest-unvalidted`

PR #594, opened 2020-04-17
fix(kustomize): allow kustomize to fetch git artifacts longer

lixin9311:
TL;DR:
When kustomization file was provided from a git repo, larger, and more historic, repositories can take a substantial amount of time to clone. `jGit` does not have sufficient functions to speedup fetching process, thus the bake process will likely to exceed the current timeout (~15s), and fail.

issue details:
https://github.com/spinnaker/spinnaker/issues/5402

work around details:
https://github.com/spinnaker/spinnaker/issues/5402#issuecomment-615172419

Overloading `downloadArtifact` with `maxRetries=7` will allow it to poll for ~1 min.

BTW, loading a configurable number from config files will be graceful.

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloader.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloader.java
@@ -24,5 +24,7 @@ import java.nio.file.Path;
 public interface ArtifactDownloader {
   InputStream downloadArtifact(Artifact artifact) throws IOException;
 
+  InputStream downloadArtifact(Artifact artifact, int maxRetries) throws IOException;
+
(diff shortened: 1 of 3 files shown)


mat-613:
Hi @ezimanyi  i've tried setting the timeout in the rosco-local.yml but the pipeline goes in timeout after 20seconds in the bake stage.

ezimanyi:
@mat-613 : Right! Of course, as orca is also waiting on this call.  Thanks for finding that.  Based on that would it make sense to close this PR and use the instructions you added to the bug?  Thanks!

lixin9311:
increase default timeout for `okHttpClient` to work around.

I thought it was async polling, but turned out being wrong.

Hillkorn:
Might help you here as well. The server is handling requests async and has its own timeouts. https://github.com/spinnaker/spinnaker/issues/6448#issuecomment-2535693583

PR #633, opened 2020-05-30
refactor(artifacts): Mutate a copy of kustomization artifact

ezimanyi:
In an upcoming kork release, the Artifact class will become immutable to reduce the confusing class of bugs where artifact mutations are accidentally done on a copy (primarily in orca). There were very few places we were mutating artifacts, and they can all be replaced by returning a new artifact with the desired changes.

The only place in rosco that is mutating artifacts is the kustomize bake. As it walks down the dependent kustomizations, it keeps track of where it is in an artifact.name field. This change just updates the code to pass a new artifact with the desired name down the recursion rather than mutating the existing one.

The one side effect here is that the input Artifact that we are baking will no longer have its name end up as the most deeply-nested kustomization that it refers to and will keep the name it started with.

I doubt the intention in any case was to have the incoming artifact be mutated in this way so that its name (but nothing else) referred to the last-processed of its nested customiztions. But as the name doesn't affect the downloading or baking logic it looks like this was not an issue (and updating this to operate on a copy and leave the original artifact as is should not have any user-visible effect).

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java
@@ -259,8 +259,12 @@ public class KustomizeTemplateUtils {
       // look like a folder then we know it should be downloaded later.
       if (isFolder(evaluate)) {
         Path tmpBase = Paths.get(FilenameUtils.normalize(base.resolve(evaluate).toString()));
-        artifact.setName(tmpBase.toString());
-        filesToDownload.addAll(getFilesFromArtifact(artifact, referenceBaseURL, tmpBase, filename));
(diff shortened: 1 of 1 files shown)


ezimanyi:
There was one more small change I needed to support encapsulating the metadata; added the commit to this PR rather than create a new one as it's not merged yet.

* refactor(artifacts): Use toBuilder rather than copy all fields

  Another change to artifacts is that we will no longer allow access to the full metadata map via .getMetadata and will instead require callers to ask for they key they want via getMetadata(String key).

  In the one call here, we were only using getMetadata to copy the data over to a new artifact; this can be replaced by using the new toBuilder function which copies over all fields. We can then set the one field that we weren't explicitly setting to the value of the original artifact.

maggieneterval:
👍  Last two commits still look good! 🎉

ezimanyi:
@ethanfrogers : Mind taking a look at this?  Thanks!

PR #642, opened 2020-06-12
feature(bakery): allow dynamic packer templates in the bakery

AbdulRahmanAlHamali:
Currently, we can only read templates that were added to rosco by halyard.

This PR allows to pass templates dynamically.

Since the `packer build` command only accepts a file, I `echo` the template, and pass that output to the command. So it looks like:

```
echo {"my_values": "5"} | packer build -
```


AbdulRahmanAlHamali:
This doesn't work, I will submit a different one that does

PR #697, opened 2020-08-25
fix(bake): Fix error when updating artifacts after bake

ezimanyi:
Fixes spinnaker/spinnaker#5998.

* test(bake): Add test exercising bake artifact handling

  There is currently no test that exercises the logic that updates the name and uuid on artifacts produced by a bake.  Add a test exercising this logic that currently fails and will be fixed in the next commit.

* fix(bake): Fix error when updating artifacts after bake

  When artifacts were made immutable, this usage was missed as it did not have tests (and was in groovy so compilation didn't catch it). Instead of mutating the artifacts, return copies with the appropriate fields changed.

### DIFF
--- a/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
+++ b/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/executor/BakePoller.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.rosco.executor
 
 import com.netflix.spectator.api.Registry
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.rosco.api.Bake
 import com.netflix.spinnaker.rosco.api.BakeRequest
 import com.netflix.spinnaker.rosco.api.BakeStatus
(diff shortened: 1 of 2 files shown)


ezimanyi:
This was a regression in 1.22 introduced by #663 (which itself was pulling in `kork` change spinnaker/kork#701).  I've added tests demonstrating the bug and the fix, so this meets the [release branch patch criteria](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria) and I'll cherry-pick this to 1.22 after it has merged.

german-muzquiz:
@ezimanyi since the issue originated in kork, do you think there may be other services affected besides rosco?

ezimanyi:
@german-muzquiz : The change in `kork` removed a few deprecated functions (setters on `Artifact`); it is possible that there could be places in other services that were using these deprecated functions.

In Java code this would be a compile time error, and in Groovy code this would cause a run-time exception, so this could only occur in Groovy code that is not exercised by tests.

The history is:
* In 1.21, I deprecated the functions in `kork` and removed all usages in the microservices (evidently missing this one)
* In 1.22, I removed the deprecated functions

I think if there are a lot more issues that arise (or any issues where it is difficult to remove the deprecated functions) we can consider reverting the `kork` change in 1.22; but I think that as long as it's just one or two small cases like this then it's easier and safer to just fix the affected microservice to no longer use the deprecated function.

german-muzquiz:
@Mergifyio backport release-1.22.x

PR #705, opened 2020-09-01
feat(provider/cf): add bake cf manifest stage

zachsmith1:
This is a new stage specifically for the cloudfoundry provider. This stage is similar to the kubernetes bake manifest stage in that its goal is to produce manifest artifacts from templated manifests and overrides/variables files. The CF CLI currently supports this exact feature, but doesn't expose it externally. Our goal with this stage is to closely mimic the functionality in the CF CLI and provide users the ability to use a templated manifest with 1-N variables files. This stage will allow users to create CF manifests on-the-fly, specific to the deployment environment, without having to manage a separate manifest for each.
Screen Shot 2020-09-01 at 1 30 39 PM

https://github.com/spinnaker/deck/pull/8529
https://github.com/spinnaker/orca/pull/3886

### DIFF
--- a/rosco-manifests/rosco-manifests.gradle
+++ b/rosco-manifests/rosco-manifests.gradle
@@ -11,6 +11,7 @@ dependencies {
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-compress:1.14"
   implementation "org.yaml:snakeyaml:1.25"
+  implementation "org.springframework.vault:spring-vault-core"
 
   implementation "com.squareup.retrofit:retrofit"
   testImplementation "org.assertj:assertj-core"
--- a/rosco-manifests/rosco-manifests.gradle
+++ b/rosco-manifests/rosco-manifests.gradle
@@ -11,7 +11,6 @@ dependencies {
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-compress:1.14"
   implementation "org.yaml:snakeyaml:1.25"
-  implementation "org.springframework.vault:spring-vault-core"
 
   implementation "com.squareup.retrofit:retrofit"
   testImplementation "org.assertj:assertj-core"
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
@@ -83,7 +83,7 @@ public class CloudFoundryBakeManifestService
     }
     vars = flatten(vars);
 
-    Set<String> matches = new HashSet<>();
+    Set<String> unresolvedKeys = new HashSet<>();
     Matcher m = Pattern.compile(pattern).matcher(manifestTemplate);
 
     while (m.find()) {
@@ -91,13 +91,13 @@ public class CloudFoundryBakeManifestService
       if (vars.get(key) != null) {
         manifestTemplate = manifestTemplate.replace(m.group(), (String) vars.get(key));
       } else {
-        matches.add(m.group());
+        unresolvedKeys.add(m.group());
       }
     }
 
-    if (matches.size() > 0) {
+    if (unresolvedKeys.size() > 0) {
       throw new IllegalArgumentException(
-          "Unable to resolve values for the following keys: \n" + String.join("\n ", matches));
+          "Unable to resolve values for the following keys: \n" + String.join("\n ", unresolvedKeys));
     }
 
     return Artifact.builder()
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2020 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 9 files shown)


rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    Artifact artifact = cloudFoundryBakeManifestService.bake(request);
+    String resolvedManifest =
+        new String(Base64.getDecoder().decode(artifact.getReference()), "UTF-8");
+    String expectedManifest = "{\"app\":\"bobservice1\",\"version\":\"v999\"}";
+    assertThat(resolvedManifest).asString().isEqualTo(expectedManifest);
+  }
+
+  @Test
+  public void shouldResolveManifestWithComplexVars() throws IOException {
```
german-muzquiz, line 114: Can you add a negative test where the overrides cannot be mapped to the template? I'm not sure what should be expected in the code, but I think end users should see a human readable error in the pipeline execution UI, so that they can fix the mapping without trying to guess what failed.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
```diff
@@ -0,0 +1,92 @@
...
+                artifactDownloader.downloadArtifact(bakeManifestRequest.getManifestTemplate()),
+                Charsets.UTF_8));
+
+    Map<String, Object> vars = new HashMap<>();
+    for (Artifact artifact : bakeManifestRequest.getVarsArtifacts()) {
+      InputStream inputStream = artifactDownloader.downloadArtifact(artifact);
+      vars.putAll(yaml.load(inputStream));
+      inputStream.close();
+    }
+    vars = JsonMapFlattener.flatten(vars);
```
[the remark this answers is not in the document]
zachsmith1: The CF native way is yaml files with key/value pairs. We use that flattener because it gives us the flattened keys for free without having to introduce that same logic.

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestServiceTest.java
zachsmith1: Currently, that feature doesn't exist. If we provided a flag like "failOnUnknownKeys" then we could accomplish this and fail the stage for unknown keys.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
```diff
@@ -0,0 +1,92 @@
...
+import org.yaml.snakeyaml.Yaml;
+
+@Component
+public class CloudFoundryBakeManifestService
+    extends BakeManifestService<CloudFoundryBakeManifestRequest> {
+
+  private static final ImmutableSet<String> supportedTemplates =
+      ImmutableSet.of(BakeManifestRequest.TemplateRenderer.CF.toString());
+  private final ArtifactDownloader artifactDownloader;
+  private final Yaml yaml = new Yaml();
```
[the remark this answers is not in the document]
zachsmith1: Good catch -- we can move this to the method body

rosco-manifests/rosco-manifests.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
   implementation "commons-io:commons-io"
   implementation "org.apache.commons:commons-compress:1.14"
   implementation "org.yaml:snakeyaml:1.25"
+  implementation "org.springframework.vault:spring-vault-core"
```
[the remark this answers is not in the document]
zachsmith1: We should just need to implement the same logic that happens in this class if its too much headache to use the import

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestService.java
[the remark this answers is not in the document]
zachsmith1: https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#variable-substitution

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/cloudfoundry/CloudFoundryBakeManifestServiceTest.java
german-muzquiz: Since we're doing here a full templating implementation, as opposed to kustomize of helm, where the replacement logic lives in those external tools, I'm trying to step back and see how a user will interact with this feature.
What should happen in this example?:
Manifest:
```
{
  "app": "((appName))",
  "version": "((someversion))"
}
```
Overrides:
```
otherAppName: test
```
I see that not all variables needed by the manifest were supplied by overrides. I'd prefer that spinnaker tells me that the baking failed because of wrong mappings, than trying to send to CF an invalid manifest that could produce unexpected results.
Another overrides example for the same manifest:
```
appName: ""test""
version: 1
```
As I read the code, it will produce invalid json:
```
{
  "app": ""test"",
  "version": "1"
}
```

zachsmith1: Both the input manifest template and the variables files are required to be yaml files. 
template.yaml
---
`application: ((someName))`

variables.yaml
---
`someName: app1`

This would resolve to:
resolved.yaml
---
`application: app1`

I'm agreeing with you that we should add a feature that would allow the user to select if the stage fails for unknown keys or not.

zachsmith1: The current logic will try to replace all variables it is aware of so that will need tweaking in order to add a flag for failing on unknown keys. It doesn't currently care if they exist or not.

german-muzquiz (approved):
Looks good, thanks for the changes!

kevinawoo (approved):

PR #733, opened 2020-10-13
fix(kustomize): Use 3.8.5 of kustomize with latest fixes to kustomize handling

jasonmcintosh:
Encountered some bugs with kustomize 3.8.1 - this bumps to 3.8.5 with the latest fixes.


jasonmcintosh:
Specific error with a kustmize build:
Error: containers path is not of type []interface{} but map[string]interface {}

dogonthehorizon (approved):

kskewes:
I think this one can be closed as we ended up merging this change into #764 .

PR #736, opened 2020-10-16
refactor(artifact downloader): Chain root cause when rethrowing exception

jludvice:
* don't hide the root cause of artifact download failure to ease up troubleshooting


ezimanyi (approved):
Looks good, thanks!  Looks like this should fix #6104.

PR #750, opened 2020-12-05
feat(exceptions): improve error handling

dbyron0:
by using the SpinnakerRetrofitErrorHandler class from kork, exceptions when rosco calls
other spinnaker microservices are more visible to the user

So, instead of seeing, for example

500

deck now shows

Status: 500, URL: http://spin-clouddriver-ro.spinnaker:7002/artifacts/fetch/, Message: Artifact credentials 'helm-account' cannot handle artifacts of type 'git/repo'

Visually, before:
image
after:
image

related to https://github.com/spinnaker/spinnaker/issues/5473


ezimanyi (approved):
💯 Thanks for picking this up!

PR #757, opened 2020-12-15
feat(provider/aws): Allow fetching of mostRecent matching AMI when searching by name

gavinbunney:
This adds a new configuration flag, `mostRecent` to `AWSVirtualizationSettings`, allowing Rosco to lookup the most recent AMI by name. This is useful to allow you to search for a given AMI series and always get the latest, e.g. use the latest ubuntu-xenial as the base ami. This feature is supported in packer via `most_recent` so this performs a similar function for Rosco.

Without this change, Rosco would simply pick the _first_ result from Clouddriver (which is typically the oldest AMI but no guarantees).


gavinbunney:
@spinnaker/reviewers

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/aws/AWSBakeHandler.groovy
```diff
@@ -231,14 +231,21 @@ public class AWSBakeHandler extends CloudProviderBakeHandler {
...
         retrySupport.retry({
           clouddriverService.findAmazonImageByName(name, account, region)
         }, 3, Duration.ofSeconds(3), false)
       }
     )
+
+    if (mostRecent) {
+      def matchingImages = images?.findAll { it?.attributes?.virtualizationType == vmType }
+      def image = matchingImages?.max {it?.attributes?.creationDate }
+      return image?.amis?.get(region)?.first()
```
th3morg, line 246: Given that line 246 and 250 are the same, would it be beneficial to reduce this to one return statement?

gavinbunney: Good idea! Fixed that up

ajordens (approved):

PR #764, opened 2021-01-21
feat(packer): Update version to support AWS assume_role

kskewes:
And other fixes and features, see: https://github.com/hashicorp/packer/blob/master/CHANGELOG.md

Tested by building `Dockerfile.slim` image, running it then doing a multi-region bake with the same custom Packer json config file as was using with Packer 1.4.5.

Closes: https://github.com/spinnaker/spinnaker/issues/6235

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,7 +1,7 @@
 FROM alpine:3.11
 LABEL maintainer="sig-platform@spinnaker.io"
 
-ENV KUSTOMIZE_VERSION=3.8.1
+ENV KUSTOMIZE_VERSION=3.8.5
 ENV PACKER_VERSION=1.6.6
 
 WORKDIR /packer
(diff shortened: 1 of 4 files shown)


Dockerfile.slim
```diff
@@ -1,12 +1,15 @@
 FROM alpine:3.11
 LABEL maintainer="sig-platform@spinnaker.io"
 
+ENV KUSTOMIZE_VERSION=3.8.1
```
jasonmcintosh, line 4: I'd HIGHLY recommend 3.8.5 - hit a bug think in 3.8.1 ( think another PR on this would have to go look)

kskewes: Thanks Jason, added commit with 3.8.5 per your PR #733.
I can drop the extra commit or we can merge this and close yours given I've also changed Dockerfile. Whatever works for everyone to get this across the line. :)

Kustomize looks to be up to 3.8.9 patch version now but I'm not familiar enough with the project to understand the changelog so probably safest to roll with 3.8.5 as you suggest. Happy to amend if someone requests though.

link108 (approved):

PR #767, opened 2021-01-25
fix(rosco): Update Hashicorp Packer version in rosco to 1.6.6

EasyAsABC123:
Using version 1.6.6 to resolve https://github.com/spinnaker/spinnaker/issues/6311 as well as get latest features of packer


avram:
See also #764 which is a version of the same

PR #818, opened 2021-10-29
Improve clouddriver exception handling

dbyron-sf:
These changes should really have been in https://github.com/spinnaker/rosco/pull/750.  That PR added detail from clouddriver, but at the expense of detail from rosco about the artifact and messages about helm charts and downloading.  This restores the info from before that PR.

image

where before this PR that was:

image


apoorvmahajan (approved):
lgtm

german-muzquiz (approved):
Thanks for this improvement!

PR #841, opened 2022-03-01
chore(ci): GHA - container image build & push

kskewes-sf:
shared:
- GHA doesn't have a repository name env var like `clouddriver` so
  add a step to extract the name from `spinnaker/clouddriver`
- Change build step to create installation distribution for use in
  following container image build steps.
- add Alpine/slim container build step
- add Ubuntu container build step

when open PR has commit pushed to fork's branch:
- build only and don't push
   - `latest{-slim|ubuntu}`
   - `{git-SHA}-{DATE-TIME}{-slim|ubuntu}`
- validates gradle build, unit tests pass and Dockerfile's build

when commit pushed to specific branches (eg: merge PR):
- build and push with branch tags:
  - `master` -> `master-latest-unvalidated-{slim|ubuntu}`
  - `master` -> `master-{git-SHA}-{DATE-TIME}-unvalidated-{slim|ubuntu}`
  - `release-*` -> `{release-*}-latest-unvalidated-{slim-ubuntu}`
  - `release-*` -> `{release-*}-{git-SHA}-{DATE-TIME}-unvalidated-{slim-ubuntu}`
- DONE: change `version-*` to `release-*` to do action on merge to release
  branches.

when push tag x.y.z (i.e: Release):
- build and push with version tag.
  - `RELEASE_VERSION` -> `{RELEASE_VERSION}-unvalidated-{slim-ubuntu}`
  - `RELEASE_VERSION*` -> `{RELEASE_VERSION}-{git-SHA}-{DATE-TIME}-unvalidated-{slim-ubuntu}`
- DONE: clarify if this is still necessary or should be revised to be
  inline with `release-x.y.z` - follow same pattern as Debian packages.

DONE: confirmed we need/want {date} in image tag name per previous CI. - Identifies distinct builds created by non-idempotent actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,10 +37,10 @@ jobs:
           file: Dockerfile.slim
           push: true
           tags: |
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-latest-unvalidated"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-latest-unvalidated"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-latest-unvalidated-slim"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-latest-unvalidated-slim"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-latest-unvalidated"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-unvalidated"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-latest-unvalidated-slim"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-unvalidated-slim"
       - name: Build and publish ubuntu container image
         uses: docker/build-push-action@v2
         with:
@@ -48,5 +48,5 @@ jobs:
           file: Dockerfile.ubuntu
           push: true
           tags: |
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-latest-unvalidated-ubuntu"
-            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-latest-unvalidated-ubuntu"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-latest-unvalidated-ubuntu"
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ env.GITHUB_REF_NAME }}-${{ github.sha }}-${{ steps.get_date.outputs.DATETIME }}-unvalidated-ubuntu"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,9 @@ jobs:
         run: echo ::set-output name=REPO::${GITHUB_REPOSITORY##*/}
       - name: Build
         run: ./gradlew build --stacktrace ${{ steps.extract_repo_name.outputs.REPO }}-web:installDist"
+      - name: Get date
+        id: get_date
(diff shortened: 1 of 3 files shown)


kskewes-sf:
TODO's/questions raised:

1. Confirm if in `build.yml` change `version-*` to `release-*` branch selector to do action on merge to release branches?
2. Confirm if still necessary to push individual versions of services or just "release" version (eg: 1.27.x). ?
3. Confirm if we need `{date}` in image tag. If we remove it then the `:latest` image is the latest and preceding images will be overwritten (different SHA though). Consider using short `SHA` instead of date? [Container registries list push date](https://console.cloud.google.com/artifacts/docker/spinnaker-community/us/docker/rosco).
4. Confirm path forward for `-validated-` tag for images. Is this doable in GHA or leave to another tool (Jenkins?).

.github/workflows/build.yml
```diff
@@ -21,5 +21,25 @@ jobs:
...
-        run: ./gradlew build --stacktrace
+        run: ./gradlew build --stacktrace ${{ steps.extract_repo_name.outputs.REPO }}-web:installDist"
+      - name: Build slim container image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile.slim
+          push: true
+          tags: |
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:spinnaker-${{ env.GITHUB_REF_NAME }}-latest-unvalidated"
```
kskewes-sf, line 36: at the top of this file should we add `release-*` to branches to match on?
```
on:
  push:
    branches:
    - master
    - release-*  # build and push "release-x-y-z-latest" on commit to release branch?
    - version-*
```

kskewes-sf, line 36: Currently:
1. all images also have a tag with `{date}` in it, eg: `spinnaker-release-1.25.x-20211208040018-ubuntu`
2. sometimes the same image (SHA) is tagged with different dates, eg: SHA `0e633dcfe<snip>` is tagged `spinnaker-release-1.25.x-20211208040018-ubuntu` and `spinnaker-release-1.25.x-20211214191852-ubuntu` amongst others.
3. it's not clear to me what timezone the date is in. This would be more obvious when GHA is doing pushes as the Action steps are publicly visible
4. having the date does give a kind of "immutable" feel, though technically there is nothing to limit an overwrite push

We could:
a) remove the `{date}` then the `:*latest*` image is the latest and preceding images will be overwritten (different SHA though).
b) consider using a short or long `SHA` instead of date to preserve individual merge commit images?

[Container registries list push date](https://console.cloud.google.com/artifacts/docker/spinnaker-community/us/docker/rosco) so it's possible to understand when the image was pushed.

.github/workflows/pr.yml
```diff
@@ -15,5 +16,25 @@ jobs:
...
+      run: "./gradlew build ${{ steps.extract_repo_name.outputs.REPO }}-web:installDist"
+    - name: Build slim container image
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: Dockerfile.slim
+        tags: |
+          "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:latest"
+          "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:latest-slim"
+          "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ github.sha }}-slim"
```
kskewes-sf, line 32: See comment above regarding using git SHA in image instead of `{date}`. Is this what we want?

kskewes-sf: This comment is on `pr.yml` but we could add a SHA like this instead of `{date}` in `build.yml`.

.github/workflows/build.yml
dbyron-sf: Depending on our naming convention, I suspect so, yes.  I figure we want every push to a release branch to get its own docker image.

dbyron-sf: I don't see `{date}` here anywhere.  Does `build-push-action` add it?  I like a git sha more than a date since it links back to the code.  But I also like a date because it helps give a natural sort order.  So my current reflex is to have both...unless we somehow tag commits with the docker tag, and then we don't need the git sha in the docker tag.

kskewes-sf: Thanks, `version-*` is unused so have switched `version-*` to `release-*` in https://github.com/spinnaker/rosco/pull/841/commits/d121964c001902f6680aa38d7a4b0bb9488f6be9

This will build and push a tag akin to a common `:latest` tag except here the tag is `spinnaker-MAJOR.MINOR.x-latest-unvalidated`. 
For example, a git push to `release-1.27.x` branch results in a new `spinnaker-1.27.x-latest-unvalidated` image being built and pushed to GAR - overwriting any existing one.

Will revisit this when understand the patch process and all of the image consumers.
If we want to preserve intermediate container images in-between official "cut" PATCH versions we could add date or commit sha to the tag.

kskewes-sf: Sorry the existing CI seems to add a `{date}` and that can be seen in https://console.cloud.google.com/artifacts/docker/spinnaker-community/us/docker/rosco.

I've added both git commit and date time to tag variations in https://github.com/spinnaker/rosco/pull/841/commits/e47dd63fb60df27701ac0b3d993a5b71afe27280

[Dockerfile's](https://github.com/spinnaker/rosco/blob/master/Dockerfile.slim#L9 ) contain non-idempotent commands like `apk update` so a single git commit could be built multiple times resulting in a different container image SHA.
If we don't have date time then we will be overwriting previous images.

Overall we might be over indexing on tags but we could pare it back later once we shake out the complete process?

.github/workflows/pr.yml
kskewes-sf: ended up with both per: https://github.com/spinnaker/rosco/pull/841#discussion_r818229396

.github/workflows/build.yml
kskewes-sf: I've added both git commit and date time to tag variations in https://github.com/spinnaker/rosco/commit/e47dd63fb60df27701ac0b3d993a5b71afe27280

.github/workflows/release.yml
```diff
@@ -39,7 +43,36 @@ jobs:
...
+      - name: Build slim container image
+        # Only run this on repositories in the 'spinnaker' org, not on forks.
+        if: startsWith(github.repository, 'spinnaker/')
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: Dockerfile.slim
+          push: true
+          tags: |
+            "${{ env.CONTAINER_REGISTRY }}/${{ steps.extract_repo_name.outputs.REPO }}:${{ steps.release_info.outputs.RELEASE_VERSION }}"
```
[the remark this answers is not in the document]
kskewes-sf: Okay, I propose that we do `validated` tagging as a second phase. WDYT?

dbyron-sf (approved):

PR #844, opened 2022-03-03
fix(publishing): don't enable maven publishing for apt release

chris-h-phillips:


chris-h-phillips:
this should prevent the google artifactregistry plugin from being applied when we are doing apt package publishing https://github.com/spinnaker/spinnaker-gradle-project/blob/master/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/artifactregistry/ArtifactRegistryPublishPlugin.groovy#L25

dbyron-sf (approved):

PR #858, opened 2022-03-10
fix(ci): GHA login to GAR and build+push for key branches

kskewes-sf:
We need to login in order to push to GAR and only specific `spinnaker`
org repositories have permission to do so.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,7 @@ jobs:
         # Only run this on repositories in the 'spinnaker' org, not on forks.
         if: startsWith(github.repository, 'spinnaker/')
         uses: docker/login-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -30,7 +30,17 @@ jobs:
       - name: Get date
         id: get_date
         run: echo ::set-output name=DATETIME::$(date --utc +'%Y%m%d%H%M')
+      - name: Login to GAR
+        # Only run this on repositories in the 'spinnaker' org, not on forks.
+        if: startsWith(github.repository, 'spinnaker/')
+        uses: docker/login-action@v1
+        with:
+          registry: us-docker.pkg.dev
+          username: _json_key
```
kskewes-sf, line 39: This is the `service account` flow defined here: https://github.com/docker/login-action#service-account-based-authentication-1

dbyron-sf (approved):

PR #883, opened 2022-05-04
fix(bake): propagate X-SPINNAKER-* headers to clouddriver

dbyron-sf:
This is an improvement on https://github.com/spinnaker/rosco/pull/748, and facilitates troubleshooting by getting e.g. X-SPINNAKER-EXECUTION-ID into the MDC and therefore log messages.  In other words it's possible to correlate more log messages with a pipeline execution now.

### DIFF
--- a/rosco-core/rosco-core.gradle
+++ b/rosco-core/rosco-core.gradle
@@ -18,7 +18,7 @@ dependencies {
   implementation "org.codehaus.groovy:groovy-all"
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "redis.clients:jedis"
-  implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0"
+  implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client"
   testImplementation "org.spockframework:spock-core"
   testImplementation "org.objenesis:objenesis"
 }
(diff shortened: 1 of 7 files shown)


rosco-core/rosco-core.gradle
```diff
@@ -18,7 +18,7 @@ dependencies {
   implementation "org.codehaus.groovy:groovy-all"
   implementation "org.springframework.boot:spring-boot-starter-web"
   implementation "redis.clients:jedis"
-  implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0"
+  implementation "com.jakewharton.retrofit:retrofit1-okhttp3-client"
```
mattgogerly, line 21: Is this specific to Rosco and not in Kork?

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImpl.java
```diff
@@ -28,13 +27,7 @@ public ArtifactDownloaderImpl(ClouddriverService clouddriverService) {
 
   public InputStream downloadArtifact(Artifact artifact) throws IOException {
     Response response =
-        retrySupport.retry(
-            () ->
-                AuthenticatedRequest.allowAnonymous(
```
mattgogerly, line 33: What's the reasoning for removing anonymous calls? Does X-SPINNAKER-USER not just get set to Anonymous?

dbyron-sf: The reasoning for adding allowAnonymous (which happened in https://github.com/spinnaker/rosco/pull/748) was to quiet down a bunch of log noise.  But now that SpinnakerRequestInterceptor is in place, there isn't any log noise to quiet down.  X-SPINNAKER-USER in the request to clouddriver is the same as it is in the request to rosco.

mattgogerly (approved):

dbyron-sf:
Thanks @mattgogerly !

PR #886, opened 2022-06-09
Arm64 support

mflis:
Changes that allow building Rosco images for both amd64 and arm64 architectures.  That's the first of many PRs to various Spinnaker Components that allow building docker images for multiple architectures.

Motivation:
Allow running & experimenting with Spinnaker locally on Macbok M1.

Changes:
- Choose binaries in docker build based on the target architecture
- bump Kustomize to a version that contains arm64 binaries
- update Github action to build multiple architectures

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - uses: actions/setup-java@v2
       - uses: actions/setup-java@v2
         with:
           java-version: 11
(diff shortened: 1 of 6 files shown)


dbyron-sf (approved):

dbyron-sf:
@mflis after merging this the build that runs on master failed: https://github.com/spinnaker/rosco/runs/6821708648?check_suite_focus=true . Mind taking a look?

.github/workflows/release.yml
```diff
@@ -22,6 +22,11 @@ jobs:
           java-version: 11
           distribution: 'zulu'
           cache: 'gradle'
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - uses: actions/setup-java@v2
```
kskewes-sf, line 29: This is already done on L20?

.github/workflows/build.yml
```diff
@@ -19,6 +19,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - uses: actions/setup-java@v2
```
kskewes-sf, line 26: Already done on line below ?

kskewes-sf (commented):
This is great, paves the way for better local development on newer arch and also experimenting with other instance types available from cloud providers.

.github/workflows/release.yml
mflis: Yeah, looks like a copy-paste error, here's a pull request with a fix: https://github.com/spinnaker/rosco/pull/887

.github/workflows/build.yml
mflis: Same as above

PR #890, opened 2022-06-21
fix(rosco): fix SSH connection issue when using RSA key, bumppacker to 1.8.1

scadu:
Fixes https://github.com/spinnaker/spinnaker/issues/6677

From upstream: 
```
This release contains the latest golang.org/x/crypto/ssh module which
implements client authentication support for signature algorithms based on
SHA-2 for use with existing RSA keys. Previously, a client would fail to
authenticate with RSA keys to servers that reject signature algorithms
based on SHA-1.
```

https://github.com/hashicorp/packer/releases/tag/v1.8.1


mattgogerly (approved):

scadu:
@mattgogerly thanks for the review! Do you know when it might be merged and released? :)

scadu:
@jasonmcintosh I'm running this setup for about two weeks now and had no issues. I swapped packer binary in the rosco container image.

scadu:
@jasonmcintosh I haven’t found any issues so far. Well, it unblocked us in fact. 
I can provide you with additional information if you would like.

jasonmcintosh:
Looking... and I need to read more on the changes... looks like a lot of the init/plugins stuff is HCL based - so... this definitely wouldn't work for some of those workflows as that'd require more code change, but this might work for basic use cases!  Would love to hear more (I'm in spinnaker slack if you want to discuss more there!)

Dockerfile.slim
```diff
@@ -1,9 +1,8 @@
 FROM alpine:3.12
 LABEL maintainer="sig-platform@spinnaker.io"
 
-ENV KUSTOMIZE_VERSION=3.8.6
```
jasonmcintosh, line 4: Removing this breaks [this I'd think.  ](https://github.com/spinnaker/rosco/blob/2c370634f065928fdabfcf2396e7031d74d2a1d6/Dockerfile.slim#L34)

Dockerfile.ubuntu
```diff
@@ -1,12 +1,13 @@
 FROM ubuntu:bionic
 LABEL maintainer="sig-platform@spinnaker.io"
 
-ENV KUSTOMIZE_VERSION=3.8.6
```
jasonmcintosh, line 4: See other comment - we shouldn't remove this.  Maybe look at bumping the 3. version of kustomize if anything, but caution flag as we've seen that cause breakage...

scadu:
> Could we change the PR name to feature(rosco): Update packer to 1.8.1 ? this looks bigger than a fix and we would like to prevent this from being back ported

Done.

jasonmcintosh (approved):

PR #899, opened 2022-08-22
fix(install): Fixed bugs in postInstall script that causes installation to fail on Ubuntu 20.04 and 22.04 LTS

ashleykleynhans:
This is a bug fix for issue [#6710](https://github.com/spinnaker/spinnaker/issues/6710) where spinnaker-rosco installation from the spinnaker-community .deb repository using Halyard fails on both Ubuntu 20.04 LTS and Ubuntu 22.04 LTS.


rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -11,17 +11,24 @@ if [ -z `getent passwd spinnaker` ]; then
   useradd --gid spinnaker spinnaker -m --home-dir /home/spinnaker
 fi
 
+create_temp_dir() {
+  PATH=$PATH:/usr/local/bin
+  TEMPDIR=$(mktemp -d /tmp/installrosco.XXXX)
```
ashleykleynhans, line 16: Added /tmp/ prefix otherwise it was creating the temp directory in the root / directory.

rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -11,17 +11,24 @@ if [ -z `getent passwd spinnaker` ]; then
   useradd --gid spinnaker spinnaker -m --home-dir /home/spinnaker
 fi
 
+create_temp_dir() {
+  PATH=$PATH:/usr/local/bin
```
ashleykleynhans, line 15: helm installation fails without this.

kskewes-sf (approved):

rosco-web/pkg_scripts/postInstall.sh
kskewes-sf: good find. 

Small nit that it isn't `temp_dir` related and is being treated as a global so could go at top of file.
I always find bash weird how you can define a variable inside a function without `local` and it becomes a global.

kskewes-sf (approved):

ashleykleynhans:
@kskewes-sf  thanks for the review, I've moved PATH to the top of the file as you suggested, and yes I agree bash is the worst language in the world, and I usually try to avoid it at all costs 😆 .

dbyron-sf (approved):

dbyron-sf:
https://github.com/Mergifyio backport release-1.27.x release-1.28.x

PR #911, opened 2022-11-02
feat(azure): Improve uniqueness of Azure Bake Key

sergio-quintero:
By adding these fields to the bake key, it will enable concurrent bakes when selecting a different image, base label or base name. 

Resolves https://github.com/spinnaker/spinnaker/issues/6738


ovidiupopa07 (approved):

nemesisOsorio (approved):
awesome

dogonthehorizon (approved):

PR #940, opened 2023-01-19
fix(rosco): additional location of spring config as optional

Luthan95:
Made additional  spring location as optional to avoid service failure in case of folder does-not exist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-korkVersion=7.159.0
+korkVersion=7.160.0
 org.gradle.parallel=true
 spinnakerGradleVersion=8.25.0
 targetJava11=true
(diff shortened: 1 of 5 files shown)


rosco-web/src/main/groovy/com/netflix/spinnaker/rosco/Main.groovy
```diff
@@ -79,7 +79,7 @@ class Main extends SpringBootServletInitializer {
     'netflix.account': '${netflix.environment}',
     'bakeAccount': '${netflix.account}',
     'netflix.stack': 'test',
-    'spring.config.additional-location': '${user.home}/.spinnaker/',
+    'spring.config.additional-location': 'optional:${user.home}/.spinnaker/',
```
dbyron-sf, line 82: Please use kork's [DefaultPropertiesBuilder](https://github.com/spinnaker/kork/blob/f86812a2cc8e74156aeeb2792a65a2c6480f2fdc/kork-config/src/main/java/com/netflix/spinnaker/kork/boot/DefaultPropertiesBuilder.java) instead.

PR #943, opened 2023-01-26
chore(dependencies): Autobump korkVersion

spinnakerbot:


dbyron-sf:
depends on https://github.com/spinnaker/kork/pull/1010

PR #947, opened 2023-02-03
chore(kustomize): add enable-helm flag for kustomize4

mazzy89:
This PR introduces the possibility to add the `enable-flag` to Kustomize.

TODO:

- [X] ~~check whether the `enable-helm` is available only for Kustomize4. In that case gate the `add` function~~ The `enable-helm` flag is available only for Kustomize4. Need to make sure to have this flag in place only for Kustomize4


dbyron-sf:
Aaah, I see https://github.com/spinnaker/spinnaker/issues/6731....

mazzy89:
@dbyron-sf many time is requested to extend upstream Helm charts with other resources. more often it is not suitable to modify the upstream helm charts. This flag empowers Kustomize to define upstream helm charts which can be consequently extended using the power of Kustomize and reducing the cost and time to implement any changes into Helm charts.

mazzy89:
@dbyron-sf not sure whether we want to unit test the following change. I haven't seen any other cases. Please advise whether it is enough. I would then make changes into Orca.

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeBakeManifestRequest.java
```diff
@@ -26,4 +26,5 @@
 public class KustomizeBakeManifestRequest extends BakeManifestRequest {
   private Artifact inputArtifact;
   private String kustomizeFilePath;
+  private boolean isEnableHelm;
```
dbyron-sf, line 29: Isn't the convention to name this enableHelm, and then the accessor is `isEnableHelm()`?
```suggestion
  private boolean enableHelm;
```

dbyron-sf:
> @dbyron-sf not sure whether we want to unit test the following change. I haven't seen any other cases. Please advise whether it is enough. I would then make changes into Orca.

I'd like to see a test that verifies that the command line arg shows up when the flag is set, and the renderer is appropriate.  Perhaps [HelmTemplateUtilsTest](https://github.com/spinnaker/rosco/blob/2ebdaddc99ecf75de498510218e1b48e077504d7/rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/helm/HelmTemplateUtilsTest.java) can provide some inspiration for a KustomizeTemplateUtilsTest?

mazzy89:
Sure. Since groovy has been marked as deprecated and no Java tests are present for kustomize, where would you want to see the tests? Should I add it anyway under groovy or create it as Java? @dbyron-sf

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeBakeManifestRequest.java
mazzy89: you are right.

dbyron-sf:
> Sure. Since groovy has been marked as deprecated and no Java tests are present for kustomize, where would you want to see the tests? Should I add it anyway under groovy or create it as Java? @dbyron-sf

Same package as the source file (`rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizeTemplateUtils.java`), and yes, in java please, so src/test/java/...

rbadillo:
@mazzy89 what changes are needed in the Orca side?

Do you have a PR we can take a look?

mazzy89:
Not yet. I do not have an Orca PR ready.

Benly-walter:
Any updates on this please?
This feature will improve our k8s deployment setup a whole lot.

mazzy89:
@Benly-walter no updates on this. We have introduced helmfile last year and helmfile can also manage kustomize. I do not see big reasons to continue with that.

Benly-walter:
> @Benly-walter no updates on this. We have introduced helmfile last year and helmfile can also manage kustomize. I do not see big reasons to continue with that.

Thanks for the update.
In our setup, we use kustomize to manage in-house apps and also for deployment of Helm apps. This means that we do not have any helm releases to maintain and would like to continue with the same approach as it works seamlessly for all our k8s clusters. For these reasons, we do not want to change to using a helmfile.

The bake manifests task works perfect for in-house apps. If the --enable-helm switch is made available for kustomization of a helm chart, it will help greatly simplify our setup. We currently have to rely on custom scripts to check if the deployments have rolled out successfully.

On a different note, do you know if it is also possible to override the below kubectl defaults which are set to false?
```
--server-side
--force-conflicts
```

jasonmcintosh:
IF someone wants to recreate this PR & add tests ... I can review!  :) Doesn't look like it'd be too hard to do...

PR #953, opened 2023-02-16
refactor(retrofit2): Use retrofit2.x client instead of retrofit1.9 for artifact fetch from clouddriver

Luthan95:
Making use of retrofit2.x with exception handling instead of retrofit 1.9, and achieve same response/exception message
RoscoClouddriverUsingRetrofit1 9
RoscoClouddriverUsingRetrofit2 x

depends on https://github.com/spinnaker/kork/pull/1004


rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -42,6 +52,25 @@ Ok3Client okClient(OkHttp3ClientConfiguration okHttpClientConfig) {
     return new Ok3Client(okHttpClientConfig.create().build());
   }
 
+  /*OkHttpClient is creatig by calling {OkHttp3ClientConfiguration}.create() which has set all the properties.
+   * As part this OkHttpClient builder  OkHttp3MetricsInterceptor added as first interceptor, but this interceptor needs spinnaker
+   * headers but interceptor in okhttp are sequential, so altering the position of interceptor and resetting to client. */
+  @Bean
+  @Primary
```
dbyron-sf, line 59: What happens if we remove this?  In general having to use `@Primary` is is a sign of problems.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -66,4 +95,23 @@ ClouddriverService clouddriverService(
         .build()
         .create(ClouddriverService.class);
   }
+
+  /*As part of retrofit2 changes, creating clouddriverservice with retrofit2 API changes */
+  @Bean
+  ClouddriverRetrofit2Service clouddriverRetrofit2Service(OkHttpClient okHttpClient) {
+    ObjectMapper objectMapper =
```
dbyron-sf, line 102: Please move this to a separate bean and share it between the retrofit and retrofit2 code.

rosco-manifests/rosco-manifests.gradle
```diff
@@ -14,11 +14,14 @@ dependencies {
...
 
   implementation "com.squareup.retrofit:retrofit"
+  implementation "com.squareup.retrofit2:retrofit"
   testImplementation "org.assertj:assertj-core"
   testImplementation "org.junit.jupiter:junit-jupiter-api"
   testImplementation "org.junit.jupiter:junit-jupiter-params"
   testImplementation "org.junit.platform:junit-platform-runner"
   testImplementation "org.junit.vintage:junit-vintage-engine"
   testImplementation "org.mockito:mockito-core"
+  testImplementation "org.mockito:mockito-inline"
```
dbyron-sf, line 24: There's a convention of using [a mockito-extensions file](https://github.com/spinnaker/rosco/blob/368f8e59ddf81eb1e8b72592fb2df209238a7f14/rosco-web/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker) instead of this dependency.  Mind doing it this way instead?

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImplTest.java
```diff
@@ -50,9 +51,13 @@ public void downloadsArtifactContent() throws IOException {
     String testContent = "abcdefg";
 
     try (ArtifactDownloaderImplTest.AutoDeletingFile file = new AutoDeletingFile()) {
-      when(clouddriverService.fetchArtifact(testArtifact))
-          .thenReturn(successfulResponse(testContent));
-      artifactDownloader.downloadArtifactToFile(testArtifact, file.path);
+      try (MockedStatic<Retrofit2SyncCall> syncCall = Mockito.mockStatic(Retrofit2SyncCall.class)) {
```
dbyron-sf, line 54: Similar mocking questions here and below.

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
Luthan95: removed OkhttpClient bean created and moved the login as part ClouddriverRetrofit2Service bean

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -51,19 +60,50 @@ RetrySupport retrySupport() {
...
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
+
+  /*As part of retrofit2 changes, creating clouddriverservice with retrofit2 API changes */
+  @Bean
+  ClouddriverRetrofit2Service clouddriverRetrofit2Service(
+      OkHttp3ClientConfiguration okHttpClientConfig,
+      SpinnakerRequestHeaderInterceptor spinnakerRequestHeaderInterceptor) {
+    /*.
+     * As part this OkHttpClient builder  OkHttp3MetricsInterceptor added as first interceptor, but this interceptor needs spinnaker
```
dbyron-sf, line 87:
```suggestion
     * As part of okHttpClientBuilder, OkHttp3MetricsInterceptor is the first interceptor, but this interceptor needs spinnaker
```

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -51,19 +60,50 @@ RetrySupport retrySupport() {
...
+    List<Interceptor> interceptors = new ArrayList<>(okHttpClientBuilder.interceptors());
+    interceptors.add(0, spinnakerRequestHeaderInterceptor);
+    interceptors.add(
+        new HttpLoggingInterceptor()
+            .setLevel(HttpLoggingInterceptor.Level.valueOf(retrofitLogLevel)));
+    okHttpClientBuilder.interceptors().removeAll(okHttpClientBuilder.interceptors());
+    okHttpClientBuilder.interceptors().addAll(interceptors);
+
+    /*
+     * ErrorHandlingExecutorCallAdapterFactory helps to achieve similar logic as retrofit and handle exceptions globally in retrofit2.
```
dbyron-sf, line 99:
```suggestion
     * ErrorHandlingExecutorCallAdapterFactory handles exceptions globally in retrofit2, similar to SpinnakerRetrofitErrorHandler with retrofit.
```

rosco-manifests/src/test/groovy/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReaderSpec.groovy
```diff
@@ -43,24 +46,23 @@ class KustomizationFileReaderSpec extends Specification {
             .type("github/file")
             .build()
 
+        mockRetrofit.when(Retrofit2SyncCall.execute(clouddriverService.fetchArtifact(baseArtifact)))
```
dbyron-sf, line 49: I'm still curious about this mocking setup.  I'd expect to see one level of mocking for `clouddriverService.fetchArtifact(baseArtifact)`, and then a second one `mockRetrofit.execute`.

rosco-manifests/src/test/groovy/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReaderSpec.groovy
```diff
@@ -47,18 +58,20 @@ class KustomizationFileReaderSpec extends Specification {
...
         k.getResources().sort() == ["deployment.yml", "service.yml"].sort()
         k.getSelfReference() == "https://api.github.com/repos/org/repo/contents/base/kustomization.yml"
     }
 
     def "getKustomization throws an exception if it can't find a valid kustomization file"() {
         given:
-        def clouddriverService = Mock(ClouddriverService) {
+        def mockCall = Mock(Call.class) {
+            execute() >> {
+                 return Response.error(404, ResponseBody.create(null, ""))
```
dbyron-sf, line 69: Why 404 here when the previous mock returned 500?

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImplTest.java
```diff
@@ -121,13 +126,8 @@ public void artifactNotFound() throws IOException {
...
   }
 
-  private Response successfulResponse(String content) {
-    return new Response(
-        "",
-        HttpStatus.OK.value(),
-        "",
-        ImmutableList.of(),
-        new TypedByteArray(null, content.getBytes()));
+  private ResponseBody successfulResponse(String content) {
```
dbyron-sf, line 129: both calls to this appear to do:
```
Response.success(200, successfulResponse(content));
```
so let's change this method to do that.

rosco-manifests/src/test/groovy/com/netflix/spinnaker/rosco/manifests/kustomize/KustomizationFileReaderSpec.groovy
Luthan95: No specific reason, I will revert it back

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
dbyron-sf: I still think some tweaks here would help.

dbyron-sf: As I think about it more though, we're going to need code like this in every service that uses retrofit2, right?  What do think about adding a method like `createForRetrofit2` (or maybe there's a better name) to [OkHttp3ClientConfiguration](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy)?

I am curious how this works with retrofit1.  Like, OkHttp3MetricsInterceptor still needs headers to do it's work (e.g. [here](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-web/src/main/java/com/netflix/spinnaker/okhttp/MetricsInterceptor.java#L120])..but how does that happen?   Like, how can we get away with the existing [create](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy#L63) method not adding [SpinnakerRequestInterceptor](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-web/src/main/groovy/com/netflix/spinnaker/okhttp/SpinnakerRequestInterceptor.groovy) ?

Luthan95: In retrofit1 SpinnakerRequestInterceptor was executed initially later the okhttp3 interceptors are executed, so headers are already present for interceptors...

Luthan95: ya, it is better to centralize, so we can use the same in other services too...

Luthan95: According to the documentation of RequestInterceptor, it will execute before every request is made.
https://www.javadoc.io/doc/com.squareup.retrofit/retrofit/1.9.0/retrofit/RequestInterceptor.html

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -51,19 +56,43 @@ RetrySupport retrySupport() {
...
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
+
+  /*As part of retrofit2 changes, creating clouddriverservice with retrofit2 API changes */
+  @Bean
+  ClouddriverRetrofit2Service clouddriverRetrofit2Service(
+      OkHttp3ClientConfiguration okHttpClientConfig) {
+
+    OkHttpClient.Builder okHttpClientBuilder = okHttpClientConfig.createForRetrofit2();
+    okHttpClientBuilder.addInterceptor(
```
dbyron-sf, line 83: Does it make sense to move this to createForRetrofit2?

Luthan95: I don't think any intermediate interceptor got into place, I will move it to createForRetrofit2 itself

rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/services/ServiceConfig.java
```diff
@@ -51,19 +54,38 @@ RetrySupport retrySupport() {
...
         .create(ClouddriverService.class);
   }
+
+  private ObjectMapper getObjectMapper() {
+    return new ObjectMapper()
+        .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_AS_NULL)
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
+
+  /*As part of retrofit2 changes, creating clouddriverservice with retrofit2 API changes */
```
dbyron-sf, line 75: Not sure this comment is helping much anymore.

dbyron-sf (approved):

PR #1019, opened 2023-09-28
fix(npe): Fixes an NPE when no overrides are set

jasonmcintosh:


dbyron-sf:
Does it make sense to add `@Nullable` [here](https://github.com/spinnaker/rosco/blob/70a06b06ef5bca01e686a940f48f767447991c74/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/BakeManifestRequest.java#L13) along with this?

jasonmcintosh:
Sure just added it :)

dbyron-sf (approved):

PR #1063, opened 2024-02-03
feat(artifact): Introduces the expansion of artifact store references in parameters 

xibz:
This commit updates the Helm util helpers that we have to include the
artifact store. This will allow us to inspect the raw strings and check
to see if we need to expand them.

Further, I opted to not utilize the deserializer this time, since this
problem seemed very specific to Helm. If that changes, we can always
just use the deserializer

Signed-off-by: benjamin-j-powell <bjp@apple.com>

### DIFF
--- a/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/HelmBakeTemplateUtils.java
+++ b/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/HelmBakeTemplateUtils.java
@@ -16,6 +16,9 @@
 
 package com.netflix.spinnaker.rosco.manifests;
 
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactReferenceURI;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStoreConfigurationProperties;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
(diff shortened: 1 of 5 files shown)


dbyron-sf:
depends on https://github.com/spinnaker/kork/pull/1145

rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/HelmBakeTemplateUtils.java
```diff
@@ -103,4 +111,34 @@ protected Path getHelmTypePathFromArtifact(
...
+   */
+  private Object expandArtifactReferenceURIs(Object value) {
+    if (artifactStore == null || !(helmConfig.isExpandOverrides() && value instanceof String)) {
+      return value;
+    }
+
+    String ref = (String) value;
+    if (ArtifactReferenceURI.is(ref)) {
+      Artifact artifact = artifactStore.get(ArtifactReferenceURI.parse(ref));
+      return artifact.getReference();
```
dbyron-sf, line 139: Tempting to add a log message here, even at debug, so there's some evidence that this is happening.  I suppose [this one](https://github.com/spinnaker/rosco/blob/be3e7923f837d12c434e0b80a9d2fe15ba75351d/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/jobs/local/JobExecutorLocal.groovy#L55) is sufficient though.

dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #1066, opened 2024-02-08
feat(aws): Upgrade packer and add SSM support (Amazon)

jervi:
This PR upgrades Packer to version v1.10.1. All the previously default providers were [moved to plugins in v1.10][1], so we also need to manually install plugins. I've opted to install Amazon, Azure and Google plugins by default, but the list can be overridden by Docker build args (or an environment variable in the case of the deb package).

This PR also installs AWS CLI 2 and the Session Manager Plugin in the Ubuntu images and in the deb package. **I opted to not install it in the Alpine images**, as the session-manager-plugin is not officially supported, and you'd need to either copy the binary or compile it from source to get it to work. I'm very open to reconsider this.

After applying this PR, you can use [SSM instead of SSH][2] to connect to instances when baking using the `amazon-ebs` provider (or friends).

[1]: https://www.hashicorp.com/blog/announcing-the-removal-of-bundled-plugins-in-hashicorp-packer
[2]: https://developer.hashicorp.com/packer/integrations/hashicorp/amazon/latest/components/builder/ebs#session-manager-plugin

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -8,7 +8,7 @@ ENV HELMFILE_VERSION=0.153.1
 
 
 ARG TARGETARCH
-ARG PLUGINS="amazon azure googlecompute"
+ARG PACKER_PLUGINS="amazon azure googlecompute"
 
 WORKDIR /packer
 
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -60,11 +60,11 @@ USER spinnaker
 
 # Install packer plugins (must be run as spinnaker user). To provide a github token (optional), run docker build with something like "--secret id=github_token,env=PACKER_GITHUB_API_TOKEN"
 RUN for plugin in $PACKER_PLUGINS ; do \
-    if [ -f /run/secrets/github_token ]; then \
-        PACKER_GITHUB_API_TOKEN=$(cat /run/secrets/github_token) packer plugins install "github.com/hashicorp/$plugin"; \
-    else \
-        packer plugins install "github.com/hashicorp/$plugin"; \
-    fi; \
+  if [ -f /run/secrets/github_token ]; then \
+    PACKER_GITHUB_API_TOKEN=$(cat /run/secrets/github_token) packer plugins install "github.com/hashicorp/$plugin"; \
+  else \
+    packer plugins install "github.com/hashicorp/$plugin"; \
+  fi; \
 done
 
 CMD ["/opt/rosco/bin/rosco"]
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -8,7 +8,7 @@ ENV HELMFILE_VERSION=0.153.1
 
 
 ARG TARGETARCH
-ARG PLUGINS="amazon azure docker googlecompute"
+ARG PLUGINS="amazon azure googlecompute"
 
 WORKDIR /packer
 
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -3,11 +3,12 @@ LABEL maintainer="sig-platform@spinnaker.io"
 
 ENV KUSTOMIZE_VERSION=3.8.6
 ENV KUSTOMIZE4_VERSION=4.5.5
-ENV PACKER_VERSION=1.8.1
+ENV PACKER_VERSION=1.10.1
 ENV HELMFILE_VERSION=0.153.1
 
 
 ARG TARGETARCH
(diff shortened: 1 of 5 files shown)


Dockerfile.slim
```diff
@@ -3,11 +3,12 @@ LABEL maintainer="sig-platform@spinnaker.io"
 
 ENV KUSTOMIZE_VERSION=3.8.6
 ENV KUSTOMIZE4_VERSION=4.5.5
-ENV PACKER_VERSION=1.8.1
+ENV PACKER_VERSION=1.10.1
 ENV HELMFILE_VERSION=0.153.1
 
 
 ARG TARGETARCH
+ARG PLUGINS="amazon azure docker googlecompute"
```
jasonmcintosh, line 11: I'm not TERRIBLY comfortable with the docker plugin - as that TYPICALLY requires host level access kinda thing in order to access the docker socket/daemon/etc.  kinda stuff and exposes things like that runc vulnerability as I recall.  

```The Docker builder must run on a machine that has Docker Engine installed. Therefore the builder only works on machines that support Docker and does not support running on a Docker remote host. You can learn about what [platforms Docker supports and how to install onto them](https://docs.docker.com/engine/installation/) in the Docker documentation.``` FYI on the docker builder.

File: rosco-web/pkg_scripts/postInstall.sh [hunk 2]
```diff
@@ -27,13 +30,18 @@ remove_temp_dir() {
 }
 
 install_packer() {
-  PACKER_VERSION="1.8.1"
-  local packer_version=$(/usr/bin/packer --version)
+  PACKER_VERSION="1.10.1"
+  local packer_version="$(/usr/bin/packer --version)"
   local packer_status=$?
   if [ $packer_status -ne 0 ] || [ "$packer_version" != "$PACKER_VERSION" ]; then
     wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip
```
jasonmcintosh, line 37: Huh just noticed this isn't including the target arch... that'd probably be ideal on the amd64 thing here...

Dockerfile.slim
jervi: I'm fine with just removing it

rosco-web/pkg_scripts/postInstall.sh
jervi: I don't think we've ever supported making deb packages for ARM, so I didn't think it was necessary. But I can add it if you think it should be here.

Dockerfile.slim
jervi: Removed (and I've updated the PR description)

rosco-web/pkg_scripts/postInstall.sh
jervi: I've added support for arm64. Not sure if the rest of the package is architecture agnostic though.

jasonmcintosh: yeah it's minor - MOST Of the docker builds TODAY should be multi-arch.  I'm seeing increased "Build for ARM!" stuff lately.

rosco-web/pkg_scripts/postInstall.sh
```diff
@@ -27,13 +35,18 @@ remove_temp_dir() {
...
 install_packer() {
-  PACKER_VERSION="1.8.1"
-  local packer_version=$(/usr/bin/packer --version)
+  PACKER_VERSION="1.10.1"
+  local packer_version="$(/usr/bin/packer --version)"
   local packer_status=$?
   if [ $packer_status -ne 0 ] || [ "$packer_version" != "$PACKER_VERSION" ]; then
-    wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_amd64.zip
+    wget https://releases.hashicorp.com/packer/${PACKER_VERSION}/packer_${PACKER_VERSION}_linux_${ARCH}.zip
     unzip -o "packer_${PACKER_VERSION}_linux_amd64.zip" -d /usr/bin
```
jasonmcintosh, line 43: need to adjust this as well to use $ARCH

jervi: 🤦‍♂️

jervi: Thanks! Fixed 👍

jasonmcintosh (approved):

PR #1073, opened 2024-02-26
test(core): add tests to verify the rosco startup with various cloud providers enabled

j-sandy:
Adding tests for AWS, Azure, Docker, Google and Oracle cloud providers to verify the rosco startup with these providers enabled.

### DIFF
--- a/rosco-core/src/test/java/com/netflix/spinnaker/rosco/providers/aws/AWSStartupTest.java
+++ b/rosco-core/src/test/java/com/netflix/spinnaker/rosco/providers/aws/AWSStartupTest.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


rosco-core/src/test/java/com/netflix/spinnaker/rosco/providers/aws/AWSStartupTest.java
```diff
@@ -0,0 +1,50 @@
...
+    })
+public class AWSStartupTest {
+
+  @Autowired AWSBakeHandler awsBakeHandler;
+
+  @Autowired RoscoAWSConfiguration.AWSBakeryDefaults awsBakeryDefaults;
+
+  @Test
+  public void startupTest() {
+    assert awsBakeHandler
```
dbyron-sf, line 45: Can you explain why this is a good startup test?  Also, I'd rather see a more test-y assertion, like:
```suggestion
   assertThat(awsBakeHandler.getMaskedPackerParameters().toString()).isEqualToIgnoringCase("[aws_access_key, aws_secret_key]");
 ```
or even better:
```
assertThat(awsBakeHandler.getMaskedPackerParameters()).isEqualTo(List.of("aws_access_key", "aws_secret_key"));
```

j-sandy: Used it as an additional marker to verify the values returned from the static class. However, it would not be required because the test is to verify the startup of rosco application with AWS provider enabled. So, removing this assert statement.

dbyron-sf (approved):

PR #1088, opened 2024-04-18
fix(install): Removed install_aws_cli2 function from Debian post installation script which installs a Debian package within a Debian package causing errors, as well as errors when an aws cli is already installed (#6945)

ashleykleynhans:
This is a fix for [#6945](https://github.com/spinnaker/spinnaker/issues/6945).

* The post installation script should not attempt to install Debian packages within a Debian package, because the dpg process is already locked by itself and results in errors.
* If a version of awscli is already installed, it also generates errors.

Therefore the entire function to install aws cli v2 has been removed, and version 2 of the aws cli can be installed manually if it is required.

### DIFF
--- a/rosco-web/pkg_scripts/postInstall.sh
+++ b/rosco-web/pkg_scripts/postInstall.sh
@@ -64,25 +64,9 @@ install_helm3() {
   mv /usr/local/bin/helm /usr/local/bin/helm3
 }
 
-install_aws_cli2() {
-  if [ "$ARCH" = "amd64" ]; then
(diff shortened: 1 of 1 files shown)


dbyron-sf:
cc: @jervi since you wrote https://github.com/spinnaker/rosco/pull/1066, but LGTM. cc: @jasonmcintosh as well since you reviewed that PR.

jervi (approved):
LGTM

dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.34.x

PR #1107, opened 2024-09-09
refactor(dependency): replace groovy coordinates during upgrade of groovy 4.x

j-sandy:
Replacing the groovy coordinates from `org.codehaus.groovy` to `org.apache.groovy` supported by groovy 4.x and above versions. While upgrading groovy 4.0.15, encounter below error during build process of rosco-core module:
```
startup failed:
/rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy: -1: Access to java.lang.Object#request_id is forbidden @ line -1, column -1.
1 error

> Task :rosco-core:compileGroovy FAILED
```
In order to make `request_id` accessible, removed `final` access specifier and updated the deprecated `ApiModelProperty.readOnly()` method with `ApiModelProperty.AccessMode()` [method](https://docs.swagger.io/swagger-core/v1.5.X/apidocs/io/swagger/annotations/ApiModelProperty.html#readOnly()) in `BakeRequest.groovy`.

### DIFF
--- a/rosco-core/rosco-core.gradle
+++ b/rosco-core/rosco-core.gradle
@@ -1,7 +1,7 @@
 dependencies {
   api "io.spinnaker.kork:kork-artifacts"
   api "io.spinnaker.kork:kork-plugins"
-  api "org.codehaus.groovy:groovy"
+  api "org.apache.groovy:groovy"
 
   implementation "com.netflix.frigga:frigga"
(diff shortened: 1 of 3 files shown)


dbyron-sf:
included in https://github.com/spinnaker/rosco/pull/1108

PR #1121, opened 2024-11-22
fix(install): Fixed packer version check for packer installation

ashleykleynhans:
Fixes the following error during post installation of the Rosco debian package:
```
Unpacking spinnaker-rosco (1.22.0) over (1.21.0) ...
Setting up spinnaker-rosco (1.22.0) ...
/var/lib/dpkg/info/spinnaker-rosco.postinst: 53: Syntax error: redirection unexpected
dpkg: error processing package spinnaker-rosco (--configure):
```

### DIFF
--- a/rosco-web/pkg_scripts/postInstall.sh
+++ b/rosco-web/pkg_scripts/postInstall.sh
@@ -39,7 +39,7 @@ install_packer() {
   local packer_version="$(/usr/bin/packer --version)"
   local packer_status=$?
 
-  if [ $packer_status -ne 0 ] || ! grep -q "$PACKER_VERSION" <<< "$packer_version" ; then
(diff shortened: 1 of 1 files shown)


jasonmcintosh (approved):
<<< SHOULD have worked but this does work.

ashleykleynhans:
> <<< SHOULD have worked but this does work.

Probably didn't work because the post init script is `#!/bin/sh` and not `#!/bin/bash`.
When I tested it locally before approving the PR, I tested with bash and not bourne shell so thats probably where the discrepancy came in, apologies for that.

ashleykleynhans:
I just tested and confirmed that `<<<` works with bash, but not with sh.

himanhsugusain:
For anyone struggling with local install , just patch `/var/lib/dpkg/info/spinnaker-rosco.postinst` locally till release.

jasonmcintosh:
@Mergifyio backport release-1.36.x

PR #1135, opened 2025-01-26
fix(helm): teach rosco to write helm overrides to a file

dbyron-sf:
to address an error in HelmTemplateUtils that occurs when the command line becomes too long due to a large number of overrides. The solution is to introduce a configurable threshold (`helm.overridesFileThreshold`). If the total length of override values exceeds this threshold, write them to a file and use the `--values` option when invoking Helm. The default behaviour (threshold of 0) retains the current approach of not using a file, regardless of the length of the overrides.

Note that there are a number of commits here, as we iterated on the fix internally.   It turns out it's super complicated to get info in a values file with the same behavior in both helm v2 and helm v3 that we get with --set/--set-string.  At this point it probably makes the most sense to look at all commits together, but I left them separate at least to start if folks are curious about the journey.


rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/HelmBakeTemplateUtils.java
```diff
@@ -126,12 +126,17 @@ protected List<String> buildOverrideList(Map<String, Object> overrides) {
         .collect(Collectors.toList());
   }
 
+  /** Accessor for whether to expand artifact reference URIs */
+  protected boolean isExpandArtifactReferenceURIs() {
+    return (artifactStore != null && helmConfig.isExpandOverrides());
```
xibz, line 131: Can you explain the use case for this a little bit? Im unsure why you want to not expand something in rosco? Or maybe Im misunderstanding how this works.

dbyron-sf: Before this change, artifact references would sometimes remain in the values file, e.g.:
```
Created overrides file at /tmp/rosco-9390928878072864271/overrides_d07566cf-e592-4b43-ad94-d947782f4a88.yml with the following contents:
---
my-override: "ref://dbyrontest/7678d0be07b31c883c436a70522c58ec0f724f1bf8bbeb293165e330f495e321"
```

dbyron-sf: The override in question was configured as:
```
"my-override": "${#stage(\"Bake (Manifest)\").outputs.resolvedExpectedArtifacts[0].boundArtifact.reference}"
```

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/helm/HelmSetArgumentParserTest.java
```diff
@@ -0,0 +1,254 @@
...
+ * behavior. Reference to Helm's Go parser tests for comparison and validation: <a
+ * href="https://github.com/helm/helm/blob/v3.14.2/pkg/strvals/parser_test.go">Helm Go Parser
+ * Tests</a>.
+ */
+public class HelmSetArgumentParserTest {
+
+  private static Stream<Arguments> helmParseArgs() {
+
+    return Stream.of(
+        Arguments.of("name1=value1", Map.of("name1", "value1"), true),
```
xibz, line 45: Can you add a float test?

xibz (changes requested):
Overall looks good. I am unsure how float values would work, so I wanted a float test, but otherwise this looks good.

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/helm/HelmSetArgumentParserTest.java
dbyron-sf: Done in https://github.com/spinnaker/rosco/pull/1135/commits/a0da916e0c79f8399d65a7d51ee7c21e81c9508c.

xibz (approved):

rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/helm/HelmSetArgumentParserTest.java
dbyron-sf: Turns out things are more complicated than we originally thought...

xibz (approved):

